PR #1, opened 2023-03-20
initial smart contract dev setup

JamesEarle:
Add initial setup for smart contract dev with required dependencies.

Include `logdown` for logging as well as `prettier` for formatting. In VS Code, you can set to format on every save by navigating to `File > Preferences > Settings` and searching for "format on save"

This requires having a formatter, so make sure you install all the dependencies with yarn first.

image


.prettierrc
```diff
@@ -0,0 +1,26 @@
...
+        "printWidth": 80,
+        "tabWidth": 2,
+        "useTabs": false,
+        "singleQuote": false,
+        "bracketSpacing": false,
+        "explicitTypes": "preserve"
+      }
+    },
+    {
+      "files": "*.ts",
```
Will-Hyde, line 15: is it worth it to run prettier on .json also?

contracts/Registry.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
```
Will-Hyde, line 2: Should we lock the version? Or are we comfortable with allowing  it to upgrade to minor releases?

Will-Hyde (approved):

contracts/Registry.sol
Whytecrowe: should be fine, but I like locking as well

.prettierrc
Whytecrowe: we can add it later if needed.

Whytecrowe (approved):

contracts/Registry.sol
JamesEarle: I will lock the version

PR #2, opened 2023-03-21
feat: New ZNS Registry for recording domain data

JamesEarle:
Registry contract that has records of domainNameHash values as that contain their owning account as well as their resolver.

image


contracts/IRegistry.sol
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * @dev Emit when a domain's record is set and all properties are modified
+   * @param owner The owner of the domain
+   * @param resolver The resolver for the domain
+   * @param domainNameHash The identifying hash of a domain's name
+   */
+  event DomainRecordSet(
+    address indexed owner,
+    IResolver indexed resolver,
+    bytes32 domainNameHash
```
RyanCross: I think we should consider also emitting the `domainName` string here. Even if we don't explicitly keep it in storage.  We need a permanent record of name -> hash in our ecosystem.

JamesEarle: @RyanCross We will be able to get the human readable name from the DomainToken contract when that's developed. It actually holds the NFT, but this contract only contains the domain name hash which we can't get the domain name from

contracts/IRegistry.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.18;
+
+import {IResolver} from "./IResolver.sol";
+
+interface IRegistry {
```
Whytecrowe: let's call it ZNSRegistry so we know which contracts are ours at all times. some of them might be implemented by other parties

Whytecrowe: same for other contracts

contracts/IRegistry.sol
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * @dev The `DomainRecord` struct is meant to hold relevant information
+   * about a domain, such as its owner and resolver.
+   */
+  struct DomainRecord {
+    address owner;
+    IResolver defaultResolver;
```
Whytecrowe: we need to store `address` type here. different Resolvers will have different Interfaces

contracts/IRegistry.sol
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * @dev Emit when a domain's record is set and all properties are modified
+   * @param owner The owner of the domain
+   * @param resolver The resolver for the domain
+   * @param domainNameHash The identifying hash of a domain's name
+   */
+  event DomainRecordSet(
+    address indexed owner,
+    IResolver indexed resolver,
```
Whytecrowe: address type here as well

contracts/IRegistry.sol
```diff
@@ -0,0 +1,130 @@
...
+  function exists(bytes32 domainNameHash) external view returns (bool);
+
+  /**
+   * @dev Set an `operator` as `allowed` to give or remove permissions for all
+   * domains owned by `msg.sender`
+   *
+   * @param operator The account to allow/disallow
+   * @param allowed The true/false value to set
+   */
+  function setDomainOperator(address operator, bool allowed) external;
```
Whytecrowe: from a recent discussion, having "domain" in the name might be confusing since we do not give it per domain, but rather per owner

contracts/IRegistry.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.18;
+
+import {IResolver} from "./IResolver.sol";
```
Whytecrowe: not needed here

contracts/Registry.sol
```diff
@@ -1,6 +1,190 @@
...
+
+    emit OperatorPermissionSet(msg.sender, operator, allowed);
+  }
+
+  /**
+   * @dev Verify if an account is an allowed operator on domains owned by `owner`
+   * @param owner Owner of the domains to be operated on
+   * @param operator Operator of modifications to the domains, if allowed
+   */
+  function isAllowedOperator(
```
Whytecrowe: we can maybe combine this with "isOwner" concept. so that it checks both in one place and we can use it as a modifier

contracts/IRegistry.sol
Whytecrowe: also, why "default" in the name?

contracts/Registry.sol
```diff
@@ -1,6 +1,190 @@
...
+
+  /**
+   * @dev Get the hash for a given domain name
+   * @param domainName The name of the domain
+   */
+  function getDomainNameHash(
+    string memory domainName
+  ) external pure returns (bytes32) {
+    require(bytes(domainName).length != 0, "ZNS: No domain");
+    return keccak256(abi.encodePacked("znsdomain", domainName));
```
Whytecrowe: "znsdomain" should be a constant in the contract that we can get from outside.
also, we might want to move this function to Registrar. please, add a TODO comment so that we can find it easily later.
also, not sure what this does exactly... :)

contracts/Registry.sol
```diff
@@ -1,6 +1,190 @@
...
+  function _isValidDomain(bytes32 domainNameHash) internal view {
+    require(domainNameHash.length != 0, "ZNS: No domain given");
+    require(_exists(domainNameHash), "ZNS: Not a domain");
+  }
+
+  /**
+   * @dev Revert if `msg.sender` is not the owner or an operator allowed by the owner
+   * @param domainNameHash The identifying hash of a domain's name
+   */
+  function _onlyDomainOwnerOrOperator(bytes32 domainNameHash) internal view {
```
Whytecrowe: why not make it a modifier?

JamesEarle: We used internal functions before instead of modifiers because they are more optimal in situations where you're constrained by contract size ([source](https://0xmacro.com/blog/solidity-gas-optimizations-cheat-sheet/)), as we were with the original registrar contract. I suppose it became a habit, but I will change to use a modifier instead.

contracts/IRegistry.sol
JamesEarle: Makes sense, but also the operator permissions are specific to domains of the owner, not the owner themselves. I'll change it but that's the thought I have

contracts/Registry.sol
JamesEarle: What you're describing is already done in the `onlyOwnerOrOperator` modifier (that was an internal function before)

contracts/Registry.sol
JamesEarle: This is what we discussed yesterday, a simple view function to return the hashed domain name if it's needed by say the DomainToken contract, where names are kept in their human readable form

Whytecrowe: DomainToken will probably NOT keep the names in human readable form. for now I see only Registrar being able to do that. even then, it will not keep them, but can help users hash from a string label. after it gets in the system past Registrar, contracts only view domains as hashes, so that they are the same across contracts.

Why do we encode with "znsdomain" string? what about the parent domain hash?
Shouldn't it be `hash(encode(hash("parentName"), stringLabel))` ?

contracts/Registry.sol
Whytecrowe: makes sense. although, modifiers are a good and convenient convention to have. imo, we should really only remove them when we are very tight on the bytecode size.

contracts/Registry.sol
JamesEarle: We don't keep a child domain -> parent domain pointer, so the parent domain would have to be a param to this func. If you think it's possible we won't keep this here at all, maybe we should just remove it?

beardhero: Following what ens does, this "znsdomain" value should be the namehash of the TLD. eg namehash("wilder.")

contracts/Registry.sol
```diff
@@ -1,6 +1,190 @@
...
+    _checkSetDomainProperty(domainNameHash);
+    _setDomainResolver(domainNameHash, resolver);
+  }
+
+  /**
+   * @dev Get the hash for a given domain name
+   * @param domainName The name of the domain
+   */
+  function getDomainNameHash(
+    string memory domainName
```
beardhero: This domainName param should not be the human-readable name. It should be the namehash of the human-readable name.
This is so that we can handle a name at any depth, without any need for pointers from child -> parent

JamesEarle: The function is to get the `domainNameHash` value of the domain name. If we expect that in the parameter, we assume the user already has that value which would render this useless anyways.

In talks with Kirill we're not sure about where a function like this will end up, but it's more likely going to be the ZNSRegistrar when that is made, maybe. We're not sure but it will be finalized before deployment.

contracts/IZNSRegistry.sol
```diff
@@ -0,0 +1,157 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.18;
+
+import {IZNSResolver} from "./IZNSResolver.sol";
```
Whytecrowe: don't think we need this here

contracts/ZNSRegistry.sol
```diff
@@ -0,0 +1,236 @@
...
+pragma solidity ^0.8.18;
+
+import {ERC1967UpgradeUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/ERC1967/ERC1967UpgradeUpgradeable.sol";
+import {IZNSRegistry} from "./IZNSRegistry.sol";
+
+contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
+  /**
+   * @dev Constant string used in the hashing of domain names
+   */
+  // What/where should this be?
```
Whytecrowe: for comments like these, could you please add TODO in the beginning. many IDEs make it very easy to look for them when doing final software review and make them a different color. otherwise we might forget about them

contracts/ZNSRegistry.sol
```diff
@@ -0,0 +1,236 @@
...
+   * @dev Creates a new domain record owned by `msg.sender`
+   * @param domainNameHash The identifying hash of a domain's name
+   * @param resolver The resolver to set
+   */
+  function createDomainRecord(bytes32 domainNameHash, address resolver) public {
+    require(
+      domainNameHash.length != 0 && domainNameHash != 0x0,
+      "ZNS: No domain given"
+    );
+    require(!_exists(domainNameHash), "ZNS: Domain exists");
```
Whytecrowe: is there a way to optimize this along with the `validDomain` modifier? the difference seems to be only the "NOT" for `_exists` function. so that the solution can be used accross the contract in the same-ish way.

contracts/ZNSRegistry.sol
```diff
@@ -0,0 +1,236 @@
...
+   * Update the domain's default resolver
+   * @param domainNameHash The identifying hash of a domain's name
+   * @param resolver The new default resolver
+   */
+  function setDomainResolver(
+    bytes32 domainNameHash,
+    address resolver
+  ) public validDomain(domainNameHash) onlyOwnerOrOperator(domainNameHash) {
+    address currentResolver = records[domainNameHash].resolver;
+    require(currentResolver != resolver, "ZNS: Same resolver");
```
Whytecrowe: these 2 lines above are not needed. extra gas for no reason, really.

contracts/ZNSRegistry.sol
```diff
@@ -0,0 +1,236 @@
...
+   */
+  function _exists(bytes32 domainNameHash) internal view returns (bool) {
+    return records[domainNameHash].owner != address(0);
+  }
+
+  /**
+   * @dev Set a domain's owner
+   * Note that we don't check for `address(0)` here. This is intentional
+   * because we are not currently allowing reselling of domains and want
+   * to enable burning them instead by transferring ownership to `address(0)`
```
Whytecrowe, line 198: good comment!

hardhat.config.ts
```diff
@@ -1,5 +1,10 @@
 import { HardhatUserConfig } from "hardhat/config";
 import "@nomicfoundation/hardhat-toolbox";
+import "@nomiclabs/hardhat-ethers";
+import "@nomicfoundation/hardhat-network-helpers";
+
+// import "@nomicfoundation/hardhat-chai-matchers";
+require("@nomicfoundation/hardhat-chai-matchers");
```
Whytecrowe: why `require` here instead of `import`?

src/deploy.ts
```diff
@@ -1,19 +1,28 @@
-import { ethers } from "hardhat";
+import * as hre from "hardhat";
+import { Registry, Registry__factory } from "../typechain";
```
Whytecrowe: is this an older typechain object? since the contract name has changed

src/deploy.ts
```diff
@@ -1,19 +1,28 @@
...
+  await registry.connect(deployer).createDomainRecord(nameHash, fakeResolver.address);
+
+  const gasA = await registry.connect(deployer).testFuncA(nameHash);
+  const gasB = await registry.connect(deployer).testFuncB(nameHash);
+
+  // With internall function
+  console.log(gasA);
+
+  // With modifier
+  console.log(gasB);
```
Whytecrowe: nice! in the future we want to create a gas test "stencil" so that we can track all the gas changes every time we change SC code and also test different ways of implementing solutions.
we will eventually be charging users a decent amount, so, ideally, we want to make transactions cheaper anywhere we can, so that it doesn't seem too expensive to use our platform.

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+    const registryFactory = new ZNSRegistry__factory(deployer);
     registry = await registryFactory.deploy();
   })
-  it ("Gets the number", async () => {
-    console.log(await registry.number());
+  // tests to verify that ownership of a domain allows modifying it
+  // test to verify that being listed as an operator allows domain modifications propery
+  // and vice versa
+
+  //  a valid operator can change the owner of a domain, is this wanted?
```
Whytecrowe: good question. let's add this to the spec.

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+  describe("Operator functionality", () => {
+    it("Returns false when an operator is not allowed by an owner", async () => {
+      await registry.connect(deployer).setOwnerOperator(operator.address, false);
+
+      const allowed = await registry.isAllowedOperator(deployer.address, operator.address);
+      expect(allowed).to.be.false;
+    });
+    it("Returns true when an operator is allowed by an owner", async () => {
```
Whytecrowe, line 53: let's add empty lines between cases for better readability

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+      await expect(tx).to.be.revertedWith("ZNS: No domain given");
+    });
+    it("Fails to set an existing domain record if a domain is given that doesn't exist", async () => {
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      const tx = registry.connect(deployer).setDomainRecord(wilderDomainHash, operator.address, operator.address);
+      await expect(tx).to.be.revertedWith("ZNS: No domain found");
+    });
+    it("Fails to set a domain record for an unowned domain", async () => {
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      await registry.connect(deployer).createDomainRecord(wilderDomainHash, mockResolver.address);
```
Whytecrowe: interesting case... Maybe we shouldn't allow to set a record without an owner. this may need to revert if owner is not specified.
thoughts?

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+
+      const tx = registry.connect(operator).setDomainRecord(wilderDomainHash, operator.address, operator.address);
+      await expect(tx).to.be.revertedWith("ZNS: Not allowed");
+    });
+    it("Allows setting domain record owner as the 0 address", async () => {
```
Whytecrowe: another test for the same case.
good idea for adding this one!
let's try and figure out if we want it to be this way. what are the possible situations where this might be beneficial and what are the ones where we do not want this to happen?

remember that we also check the existence of the domain through the `owner` field. in this case our checks will show that a domain does not exist, but it will exist with the possible resolver set.

contracts/ZNSRegistry.sol
```diff
@@ -0,0 +1,236 @@
...
+  function setDomainRecord(
+    bytes32 domainNameHash,
+    address owner,
+    address resolver
+  ) public validDomain(domainNameHash) onlyOwnerOrOperator(domainNameHash) {
+    require(
+      records[domainNameHash].owner != owner ||
+        records[domainNameHash].resolver != resolver,
+      "ZNS: No record change"
+    );
```
Whytecrowe: another seemingly unnecessary check.

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+
+      expect(allowed).to.be.false;
+    });
+    it("Permits an allowed operator to modify a domain record", async () => {
+      await registry.connect(deployer).setOwnerOperator(operator.address, true);
+
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      await registry.connect(deployer).createDomainRecord(wilderDomainHash, mockResolver.address);
+
+      await registry.connect(operator).setDomainResolver(wilderDomainHash, operator.address);
```
Whytecrowe: why do we have both of these functions present (create... and set...) on the contact? they seem to be doing the same thing. I think, one should be enough.

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+
+      const owner = await registry.getDomainOwner(wilderDomainHash);
+      expect(owner).to.eq(hre.ethers.constants.AddressZero);
+    });
+    it("Fails to set domain record if the resolver is the 0 address", async () => {
+      // should set to a default register? or just fail?
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      await registry.connect(deployer).createDomainRecord(wilderDomainHash, mockResolver.address);
+
+      const tx = registry.connect(deployer).setDomainRecord(wilderDomainHash, operator.address, hre.ethers.constants.AddressZero);
```
Whytecrowe: so, we are allowing to set `owner` as 0, but not `resolver`. why?
same situation as the above comment.

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+      await expect(tx).to.be.revertedWith("ZNS: Zero address");
+    });
+    it("Fails to set the domain owner to the current owner", async () => {
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      await registry.connect(deployer).createDomainRecord(wilderDomainHash, mockResolver.address);
+
+      const tx = registry.connect(deployer).setDomainOwner(wilderDomainHash, deployer.address);
+      await expect(tx).to.be.revertedWith("ZNS: Same owner");
+    });
```
Whytecrowe: don't think we need this check at all, as said in one of the comments above. why is this beneficial, you think?

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+    it("Fails to set the domain resolver to the current resolver", async () => {
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      await registry.connect(deployer).createDomainRecord(wilderDomainHash, mockResolver.address);
+
+      const tx = registry.connect(deployer).setDomainResolver(wilderDomainHash, mockResolver.address);
+      await expect(tx).to.be.revertedWith("ZNS: Same resolver");
+    });
```
Whytecrowe: same here

test/registry.test.ts
```diff
@@ -1,19 +1,255 @@
...
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+      await registry.connect(deployer).createDomainRecord(wilderDomainHash, mockResolver.address);
+
+      const owner = await registry.getDomainOwner(wilderDomainHash);
+      expect(owner).to.eq(deployer.address);
+    });
+    it("Returns a domain owner of 0 if that domain doesn't exist", async () => {
+      const wilderDomainHash = hre.ethers.utils.id("wilder");
+
+      // We have not called to create the domain record, so it won't exist
```
Whytecrowe: or we have a domain record set with 0 as owner and a valid resolver address...

Whytecrowe (changes requested):
Amazing start!
Let's discuss the comments and iron out the things left.

hardhat.config.ts
JamesEarle: Just forgot to change back to the above commented import. I was having trouble getting this working right at first and wanted to make sure I was following the docs exactly

src/deploy.ts
JamesEarle: updated

test/registry.test.ts
JamesEarle: the `onlyOwnerOrOperator` modifier restricts other functions to only be executable by an owner or an operator, but to create a new domain it's not possible there is an owner yet because it is new. So the second function.

test/registry.test.ts
JamesEarle: setting owner as 0 burns the domain, but setting resolver to 0 for a valid domain would just break it, presumably. Do you disagree?

test/registry.test.ts
JamesEarle: It would be a wasted, unnecessary tx if you were to submit `setDomainOwner` with the value of the current owner, even if very unlikely. So this is to avoid allowing users to make that mistake

test/registry.test.ts
JamesEarle: @Whytecrowe I don't understand what you mean here. If changing `owner` to `0` on a domain is burning it in the system, how can it be considered here?

Whytecrowe (approved):

PR #3, opened 2023-03-23
[WIP] Namehash example using ethers.util.namehash

beardhero:
An example of the namehash process - starting with a human-readable name and hashing it for preparation to send to contract.

Note that the Namehash.sol contract is just a test, it doesn't represent how we'll use the namehashes on-contract (this is required in the registry).


test/registry.test.ts
```diff
@@ -67,15 +67,45 @@ describe("ZNSRegistry Tests", () => {
...
+      const labelhash = hre.ethers.utils.keccak256(hre.ethers.utils.toUtf8Bytes(label));
+      const domainId = hre.ethers.utils.keccak256(hre.ethers.utils.solidityPack(['bytes32', 'bytes32'], [wilderDomainHash, labelhash]));
+
+      // Successfully gets a domain record
+      const record = await registry.getDomainRecord(domainId);
+      expect(record.owner).to.eq(deployer.address);
+      expect(record.resolver).to.eq(mockResolver.address);
+    });
+    it("Successfully creates a 3LD record", async () => {
+      // the `utils.id` function returns the keccak256 hash of the given string
```
Whytecrowe, line 95: wrong function in the comment. and also above in another test

yarn-error.log
Whytecrowe, line 1: what is this log and why is it here?

PR #4, opened 2023-03-25
ZNSDomainToken implementation

Will-Hyde:
Adding the basic implementation for the ZNS Domain Token

Right now it exposes 2 methods:
register: Mints a new domain for a specified tokenID and sends it to a given address. This function utilizes _safemint
revoke: Burns the token with a given tokenId. This method currently requires the burner to be the owner of the token

As it stands now, the ZnsDomainToken is agnostic of other contracts in our infrastructure, and does not make any external calls


contracts/ZNSDomainToken.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import {Counters} from "@openzeppelin/contracts/utils/Counters.sol"; //For Token Ids
+import {IZNSDomainToken} from "./IZNSDomainToken.sol";
+import {IZNSRegistry} from "./IZNSRegistry.sol";
+
+//TODO: Add Documentation and Event Emission
+contract ZNSDomainToken is IZNSDomainToken, ERC721 {
```
Whytecrowe: Let's always inherit own interfaces last in the chain.
Inheritance in Solidity influences storage slot allocation and a standard way of inheriting is a good thing to have.

contracts/ZNSDomainToken.sol
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  //TODO: Add Access Control
+  function register(
+    address to,
+    uint256 tokenId,
+    bytes32 domainNameHash,
+    address resolver
+  ) external {
+    _safeMint(to, tokenId);
```
Whytecrowe, line 16: this order of operations is vulnerable to reentrancy attacks.
we usually want to write to storage first and mint/transfer after that, at the end - "checks-effects-interactions pattern"

contracts/ZNSDomainToken.sol
```diff
@@ -0,0 +1,63 @@
...
+  //TODO: Add Access Control
+  function setZNSRegistry(IZNSRegistry ZNSRegistry_) external override {
+    ZNSRegistry = ZNSRegistry_;
+    emit ZNSRegistrychanged(address(ZNSRegistry_));
+  }
+
+  //TODO: Add Access Control
+  function register(
+    address to,
+    uint256 tokenId,
```
Whytecrowe: it might be a good idea to generate tokenId here in this function, instead of expecting it to be an arg

contracts/ZNSDomainToken.sol
```diff
@@ -0,0 +1,63 @@
...
+    bytes32 domainNameHash,
+    address resolver
+  ) external {
+    _safeMint(to, tokenId);
+    ZNSRegistry.createDomainRecord(domainNameHash, resolver);
```
Whytecrowe: this comment is for @JamesEarle as well.

ZNSRegistry will set this contract as `owner` in ZNSRegistry after this call, because it uses `msg.sender` to set the `owner`.
need to iron this out after James changes the ZNSRegistry code.

contracts/ZNSDomainToken.sol
Will-Hyde: Updated ordering

contracts/ZNSDomainToken.sol
Will-Hyde: Per our discussion, It seems like the Registrar will pass in the tokenId to this function

contracts/ZNSDomainToken.sol
Will-Hyde: Removed reference to registry, since its not needed

contracts/ZNSDomainToken.sol
Will-Hyde: Removed other statements, now its purely calling _safemint

contracts/ZNSDomainToken.sol
```diff
@@ -0,0 +1,26 @@
...
+  constructor() ERC721("ZNSDomainToken", "ZDT") {}
+
+  //Mints a token with a specified tokenId, using _safeMint, and sends it to the given address
+  //TODO: Add Access Control
+  function register(address to, uint256 tokenId) external {
+    _safeMint(to, tokenId);
+  }
+
+  //Burns the token with the specified tokenId
+  //TODO: Add Access Control, replace require to also allow registry to revoke
```
JamesEarle: Just calling this TODO to attention. It should be addressed or removed if its outdated

contracts/ZNSDomainToken.sol
```diff
@@ -0,0 +1,26 @@
...
+pragma solidity ^0.8.18;
+
+import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import {IZNSDomainToken} from "./IZNSDomainToken.sol";
+
+//TODO: Add Access Control
+contract ZNSDomainToken is ERC721, IZNSDomainToken {
+  constructor() ERC721("ZNSDomainToken", "ZDT") {}
+
+  //Mints a token with a specified tokenId, using _safeMint, and sends it to the given address
```
JamesEarle: Using [natspec](https://docs.soliditylang.org/en/v0.8.17/natspec-format.html) style comments will provide annotations for users of the contract and also will make our code better for any devs reading it.

hardhat.config.ts
```diff
@@ -26,7 +28,43 @@ const config: HardhatUserConfig = {
   },
   mocha: {
     timeout: 40000
-  }
+  },
+  networks: {
+    hardhat: {
+      accounts: [
```
JamesEarle: Remove use of any real accounts (even testnet accounts) and forking for hardhat

hardhat.config.ts
```diff
@@ -26,7 +28,43 @@ const config: HardhatUserConfig = {
...
+      accounts: [`${process.env.MAINNET_PRIVATE_KEY}`],
+      url: `https://mainnet.infura.io/v3/97e75e0bbc6a4419a5dd7fe4a518b917`,
+      gasPrice: 80000000000,
+    },
+    goerli: {
+      accounts: [`${process.env.TESTNET_PRIVATE_KEY}`],
+      url: "https://goerli.infura.io/v3/77c3d733140f4c12a77699e24cb30c27",
+      timeout: 10000000
+    },
+    localhost: {
```
JamesEarle: We don't use a local network. Probably best just to remove this

test/DomainToken.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const TokenSymbol = "ZDT";
+
+    let deployer: SignerWithAddress
+    let caller: SignerWithAddress;
+    let callerAddress = "0x628016c31c2EC22fdd901ba6525A5F75B58A0326";
+    let domainToken: ZNSDomainToken;
+
+    beforeEach(async () => {
+        [deployer] = await hre.ethers.getSigners();
+        const deomainTokenFactory = new ZNSDomainToken__factory(deployer);
```
JamesEarle: nit: fix the spelling on `domain`

test/DomainToken.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+        domainToken = await deomainTokenFactory.deploy();
+        caller = await impersonateAddressWithBalance(callerAddress);
```
JamesEarle: We don't use this functionality for any reason, you could remove it and instead declare the caller as the second caught array entity from the `hre.ethers.getSigners()` call above and it would be the same.

Let's confirm whether or we will use any payable functions with these contracts. Previously we've done it with the sales only. If we don't need it, we can delete the entire helper.

test/DomainToken.test.ts
```diff
@@ -0,0 +1,75 @@
+import * as hre from "hardhat";
```
JamesEarle: nit: You'll see that in my branch `feat/znsRegistry` there is a `.prettierrc` and in my workspace I've enabled `format on save`. If we all do this it will allow a lot more consistency between us about the smaller stuff like indentation size where this file is 4 but others are 2. 

If you have opinions on what things should or should not be formatted a certain way we can discuss it

contracts/ZNSDomainToken.sol
Will-Hyde: Its marked as todo, because future stories will flesh it out

JamesEarle: Yeah I saw others after writing this and assumed as much. I initially thought it was left over and was not completed.

contracts/ZNSDomainToken.sol
Whytecrowe: yes. please, use NatSpec comments

test/DomainToken.test.ts
Whytecrowe: I agree to this. when the code is formatted the same way everywhere, it helps everyone

test/DomainToken.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        domainToken = await deomainTokenFactory.deploy();
+        caller = await impersonateAddressWithBalance(callerAddress);
+    })
+
+    describe("External functions", () => {
+        it("Registers a token", async () => {
+            const tx = await domainToken.connect(deployer).register(callerAddress, ethers.BigNumber.from("1"));
+            const receipt = await tx.wait(0);
+            //Verify Transfer event is emitted
+            expect(receipt.events?.[0].event).to.eq("Transfer");
```
Whytecrowe: if we are checking events, we should check their props as well (from, to, amount, etc.)

test/DomainToken.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        [deployer] = await hre.ethers.getSigners();
+        const deomainTokenFactory = new ZNSDomainToken__factory(deployer);
+
+        domainToken = await deomainTokenFactory.deploy();
+        caller = await impersonateAddressWithBalance(callerAddress);
+    })
+
+    describe("External functions", () => {
+        it("Registers a token", async () => {
+            const tx = await domainToken.connect(deployer).register(callerAddress, ethers.BigNumber.from("1"));
```
Whytecrowe: for tokenId checks in tests, let's use something better then "1".
there's a billion ways we can have a value of "1" anywhere. while this is somewhat satisfactory here, when the system grows, we might miss errors there for some other reason. we want our tests to be resilient and accurate over time

test/DomainToken.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            const receipt = await tx.wait(0);
+            //Verify Transfer event is emitted
+            expect(receipt.events?.[0].event).to.eq("Transfer");
+            let balance = await domainToken.balanceOf(callerAddress);
+            expect(balance).equals(ethers.BigNumber.from(1));
```
Whytecrowe: for 721 checks it's usually better to use tokenId ownership check rather then total balance check. again, for more accurate testing.

contracts/ZNSDomainToken.sol
Will-Hyde: Adding Nat spec comment structure

test/DomainToken.test.ts
Will-Hyde: Updated my format settings to use 2 spaces

test/DomainToken.test.ts
Will-Hyde: Verifying Destination and tokenId in these events now

test/DomainToken.test.ts
Will-Hyde: Im not sure I understand the concern here. Are you asking for a number other than 1? Or should I set this tokenId to a re-usable constant variable so that there isnt confusion within a test method?

Unit test states are not saved between tests, so it shouldnt matter what value is used

Will-Hyde: For now Ive set it to a reusable variable rather than relying on magic strings

test/DomainToken.test.ts
Will-Hyde: Using ownerOf method now to establish ownership

hardhat.config.ts
Will-Hyde: Removed localhost

hardhat.config.ts
Will-Hyde: Removed references to account numbers

test/DomainToken.test.ts
Will-Hyde: Fixed spelling

test/DomainToken.test.ts
Will-Hyde: Removed helper method and replaced it with the 2nd address from the getsigners() method

Whytecrowe (approved):

PR #5, opened 2023-03-29
[Prototype-WIP]:  Domain Distribution flows (do not merge!)

Whytecrowe:


contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -0,0 +1,164 @@
...
+        // set data on Registry and Resolver storage
+        _setDomainData(domainHash, msg.sender, resolver, domainAddress);
+
+        emit RootDomainRegistered(domainHash, name, msg.sender);
+
+        return domainHash;
+    }
+
+    function approveSubdomain(bytes32 parentHash, address ownerCandidate) external onlyOwner(parentHash) {
+        subdomainApprovals[parentHash] = ownerCandidate;
```
Whytecrowe: Do we actually want to approve a single purchase here? meaning, we will check the validity when buying, then reset this mapping value to false. do we need this?...

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -0,0 +1,164 @@
...
+
+            registerFor = msg.sender;
+        }
+
+        bytes32 domainHash = hashWithParent(parentHash, name);
+
+        // TODO: do we have burning here or just for Root Domains?
+        // we are always charging the caller here
+        // RDO Registrar if present or direct buyer/caller if no RDO Registrar
+        znsTreasury.stakeForDomain(domainHash, name, msg.sender, false);
```
Whytecrowe: needs to be changed to `true` since we will charge the fee for subdomains as well

.prettierrc
```diff
@@ -4,22 +4,22 @@
       "files": "*.sol",
       "options": {
         "printWidth": 80,
-        "tabWidth": 2,
+        "tabWidth": 4,
```
JamesEarle: DoryEyesGooglyEyesGIF

PR #6, opened 2023-03-31
Address resolver

beardhero:
No upgradeability.

Branched from feat/namehash.


JamesEarle:
Please merge latest on master and resolve the existing conflicts.

Also in the future please add reviewers so that we get notifcations when you make a PR.

contracts/ZNSRegistry.sol
```diff
@@ -99,6 +99,27 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
 
     emit DomainRecordCreated(msg.sender, resolver, domainNameHash);
   }
+  /**
+   * @dev Creates a new domain record owned by `msg.sender`
+   * @param label The human readable, new label being registered eg "citizen"
+   * @param parentHash The namehash of the full parent tree eg "wilder.world"
+   * @param resolver The resolver to set
+  */
+  function createSubDomainRecord(string memory label, bytes32 parentHash, address resolver) public {
```
JamesEarle: Why is this function here? This looks repetitive, and also isn't used in your work creating the address resolver? The `setSubdomainRecord` function would cover this, with the small added detail of using an `address owner` param instead of `msg.sender`

yarn-error.log
JamesEarle, line 1: Please remove this file from the commit

contracts/ZNSAddressResolver.sol
JamesEarle, line 1: Be sure to run the formatter on everything

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./IZNSAddressResolver.sol";
+
+contract ZNSAddressResolver is IZNSAddressResolver{
+    bytes4 internal constant RESOLVER_INTERFACE_ID = 0xe25f0a33;
```
JamesEarle: Please always add comments for numbers like this. Where did it come from? What does it do? Why do we need it?

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,43 @@
...
+     * @param domainNameHash The identifying hash of a domain's name
+     * @param newAddress The new domain owner
+     */
+    function setAddress(bytes32 domainNameHash, address newAddress) external onlyRegistry{
+        addressOf[domainNameHash] = newAddress;
+
+        emit AddressSet(domainNameHash, newAddress);
+    }
+
+    function supportsInterface(bytes4 interfaceId)external view returns (bool){
```
JamesEarle: Can you explain what this is for? And the above pre-calculated interface ID?


Can be made `pure`, also please use natspec formatting like above for comments.

contracts/IZNSAddressResolver.sol
JamesEarle, line 1: Dev tip: There is already a `.prettierrc` in the repo, enable "format on save" for consistency with other code in repo automatically

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,43 @@
...
+        return addressOf[domainNameHash];
+    }
+
+    /**
+     * @dev Sets the address of a domain name hash, only registry
+     * @param domainNameHash The identifying hash of a domain's name
+     * @param newAddress The new domain owner
+     */
+    function setAddress(bytes32 domainNameHash, address newAddress) external onlyRegistry{
+        addressOf[domainNameHash] = newAddress;
```
JamesEarle: Do we allow setting to `address(0)`?

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./IZNSAddressResolver.sol";
+
+contract ZNSAddressResolver is IZNSAddressResolver{
+    bytes4 internal constant RESOLVER_INTERFACE_ID = 0xe25f0a33;
+    address public registry;
+    mapping(bytes32 => address) public addressOf;
```
JamesEarle: Additionally just add a comment for each state variable too. It can be one line and short, but it should just quickly explain what it logically represents and how it's used

test/namehash.test.ts
```diff
@@ -0,0 +1,16 @@
+const { expect } = require("chai");
+import * as hre from "hardhat";
+//import { Namehash, Namehash__factory } from "../typechain";
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+
+describe("Namehash", function () {
+  let deployer: SignerWithAddress
+  
+  it("Should normalize the name", async function () {
```
JamesEarle: Looks like this is just testing ethers functionality, and I think we should only be testing our own created functionality, not a public and released 3rd party package. They already would (read: should) have much more thorough testing anyways.

IMO this file can be removed

test/registry.test.ts
```diff
@@ -67,15 +67,45 @@ describe("ZNSRegistry Tests", () => {
...
       // Successfully gets a domain record
       const record = await registry.getDomainRecord(wilderDomainHash);
       expect(record.owner).to.eq(deployer.address);
       expect(record.resolver).to.eq(mockResolver.address);
     });
+    it("Successfully creates a 2LD record", async () => {
+      // the `utils.id` function returns the hash of the given string
+      const wilderDomainHash = hre.ethers.utils.namehash("wilder");
+      const label = "world";
+      await registry.connect(deployer).createSubDomainRecord(label, wilderDomainHash, mockResolver.address);
```
JamesEarle: Please pull latest from master and merge. This function should be the `setSubdomainRecord` function and these scenarios are already tested

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,43 @@
...
+    function getAddress(bytes32 domainNameHash) external view returns(address){
+        return addressOf[domainNameHash];
+    }
+
+    /**
+     * @dev Sets the address of a domain name hash, only registry
+     * @param domainNameHash The identifying hash of a domain's name
+     * @param newAddress The new domain owner
+     */
+    function setAddress(bytes32 domainNameHash, address newAddress) external onlyRegistry{
```
JamesEarle: Have we confirmed that this is for sure `onlyRegistry`?

Does a request for the resolution of a token to it's mapped address always come from the Registrar? Does the DomainToken contract ever request it? And if so, does it request it directly? Or does it request it from the Registrar, which then requests it from the resolver?

contracts/ZNSAddressResolver.sol
beardhero: Unspecified, as of now this contract has no reason to restrict any use, the permitted setter can restrict appropriately.

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./IZNSAddressResolver.sol";
+
+contract ZNSAddressResolver is IZNSAddressResolver {
+  bytes4 internal constant RESOLVER_INTERFACE_ID = 0xe25f0a33; /// @notice pre-calculated erc-165 interface ID
```
Whytecrowe: let's make it hash at compile time. this way we have visibility on what that interface is exactly. here I am not sure, nor will other devs be

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./IZNSAddressResolver.sol";
+
+contract ZNSAddressResolver is IZNSAddressResolver {
+  bytes4 internal constant RESOLVER_INTERFACE_ID = 0xe25f0a33; /// @notice pre-calculated erc-165 interface ID
+  address public registry; /// @notice Address of ZNSRegistry for set permission
+  mapping(bytes32 => address) public addressOf; /// @notice mapping of domain hash to owning address
```
Whytecrowe: if we have a specific getter, let's make this private.

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * @dev Sets the address of a domain name hash, only registry
+   * @param domainNameHash The identifying hash of a domain's name
+   * @param newAddress The new domain owner
+   */
+  function setAddress(
+    bytes32 domainNameHash,
+    address newAddress
+  ) external onlyRegistry {
```
Whytecrowe: instead of guarding it for Registry only, it should be done for an authorized user. domain owners or their authorized parties should be able to re-set this at any time. Registry will only do this once upon registering, but the owner might want to change it to another address at any point.

test/ZNSAddressResolver.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    it("Should resolve address correctly", async function () {
+        const domainNameHash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes("wilder"));
+        await znsAddressResolver.connect(registry).setAddress(domainNameHash, addr1.address);
+
+        const resolvedAddress = await znsAddressResolver.getAddress(domainNameHash);
+        expect(resolvedAddress).to.equal(addr1.address);
+    });
+
+    it("Should support resolver interface ID", async function () {
+        const supported = await znsAddressResolver.supportsInterface("0xe25f0a33");
```
Whytecrowe: again here. what is this hash value? where do we get it from? how do we know if this test is right if we have no idea what we are hashing? :)

test/ZNSAddressResolver.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        expect(supported).to.be.true;
+    });
+
+    it("Should not support other interface IDs", async function () {
+        const notSupported = await znsAddressResolver.supportsInterface("0xffffffff");
+        expect(notSupported).to.be.false;
+    });
+});
```
Whytecrowe, line 76: missing events tests.

Whytecrowe (approved):
please, address comments, merge `master` into this branch and possibly remove the changes from `feat/namehash`. that branch was not merged and doesn't seem to have necessary code for this PR.
do not branch from `feat/namehash` in the future.

Whytecrowe (changes requested):
see the message above

contracts/ZNSAddressResolver.sol
RyanCross: Consider extending `AccessControlUpgradeable` to accomplish the above.    

@Whytecrowe @beardhero As far as roles themselves. Are there any design requirements for this yet?  Where in the system does one become an authorized party? 

I believe AccessControl comes with a built-in-defined ADMIN which could be defaulted to the owner, but we can additionally define more roles.

JamesEarle: @RyanCross might be a good idea, but I would say only add this if we need more than one role. If we only need authorized vs. not authorized, it would be excessive.

contracts/IZNSDomainToken.sol
JamesEarle, line 1: Any idea why it shows an entire deletion and addition for each file like this? Kind of interferes reviewing the real changes

contracts/IZNSResolver.sol
JamesEarle, line 1: This file should be deleted if we renamed to `IZNSAddressResolver` and the interface for that file exists now

test/ZNSAddressResolver.test.ts
beardhero: the event was already tested during the setAddress test

contracts/ZNSAddressResolver.sol
beardhero: calculateSelector() check added

contracts/ZNSAddressResolver.sol
beardhero: I added a modifier onlyOwnerOrOperator to the resolver.
It had to do two checks getDomainRecord and isAllowedOperator.
Consider adding a single check on the registry for onlyOwnerOrOperator.

contracts/IZNSDomainToken.sol
beardhero: Whoops! Real sorry about that.
It was due to my dev setup and is resolved for the future now, but this PR is still messy.

JamesEarle:
If the other one that I see in PRs view is meant to replace this now, please mark this closed

PR #8, opened 2023-04-05
feat: (MUD-211) TS and Solidity Linting

Whytecrowe:
- Uses of `eslint-config-cpt` module and extends it for linting TS code with some changes and reformatting of the current code in the repo.
- Uses `solhint` to catch style, best practice, security errors in Solidity code with some changes and reformatting of the current code in the repo.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,9 @@
+{
+  "root": true,
+  "overrides": [
+    {
+      "files": ["src/**/*.ts", "test/**/*.ts"],
+      "extends": ["@zero-tech/eslint-config-cpt/node-ts/.eslintrc.json"]
+    }
+  ]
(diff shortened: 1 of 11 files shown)


beardhero (approved):

contracts/IZNSResolver.sol
beardhero, line 1: delete

.eslintrc
```diff
@@ -0,0 +1,15 @@
+{
```
RyanCross, line 1: highly recommend adding support for:
https://typescript-eslint.io/rules/no-floating-promises/

RyanCross:
i recommend installing and configuring https://typescript-eslint.io/ so we get real typescript support

PR #9, opened 2023-04-06
Mud 188/address resolver

beardhero:
Setter permission is now the onlyOwnerOrOperator() condition on the registry.
To get that same condition on the resolver, I had to call both getDomainRecord() and isAllowedOperator.
We could reduce that to one call if the registry had a function that is just modified with onlyOwnerOrOperator()

AddressResolver tests include setting up a registry, so they could be moved into the registry tests, replacing the mockResolver made there.

### DIFF
--- a/contracts/ZNSAddressResolver.sol
+++ b/contracts/ZNSAddressResolver.sol
@@ -5,14 +5,12 @@ import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 import "./IZNSAddressResolver.sol";
 import "./IZNSRegistry.sol";
 
-contract ZNSAddressResolver is IZNSAddressResolver {
-  bytes4 internal resolverInterfaceId; /// @notice erc-165 interface ID, calculated in constructor
+contract ZNSAddressResolver is ERC165, IZNSAddressResolver {
   IZNSRegistry public registry; /// @notice IZNSRegistry
   mapping(bytes32 => address) private addressOf; /// @notice mapping of domain hash to address
 
   constructor(IZNSRegistry _registry) {
     registry = _registry;
-    resolverInterfaceId = calculateSelector();
   }
 
   /**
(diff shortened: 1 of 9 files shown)


contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * @dev Sets the address of a domain name hash, only registry
+   * @param domainNameHash The identifying hash of a domain's name
+   * @param newAddress The new domain owner
+   */
+  function setAddress(
+    bytes32 domainNameHash,
+    address newAddress
+  ) external onlyOwnerOrOperator(domainNameHash) {
+    addressOf[domainNameHash] = newAddress;
```
Whytecrowe, line 59: should require address not to be zero.

test/ZNSAddressResolver.test.ts
```diff
@@ -0,0 +1,79 @@
+const { expect } = require("chai");
```
Whytecrowe: this should yell at you when you merge with the new eslint rules on master :)

contracts/ZNSAddressResolver.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./IZNSAddressResolver.sol";
+import "./IZNSRegistry.sol";
+
+contract ZNSAddressResolver is ERC165, IZNSAddressResolver {
+  IZNSRegistry public registry; /// @notice IZNSRegistry
+  mapping(bytes32 => address) private addressOf; /// @notice mapping of domain hash to address
```
Whytecrowe: with 0.8.18 we can add var names for mapping members. I've added this to the solhint config, so it will tell you about it when you lint after merging `master` into your branch.

Whytecrowe (approved):
Good stuff!
Please, merge `master` into your branch, run TS and Solidity linters and fix errors if any.
After that we are good to merge

contracts/ZNSAddressResolver.sol
beardhero: Ok not sure what they're for but added the names domainNameHash and resolvedAddress

JamesEarle (approved):
LGTM 🚢 

Just do the several minor things required to get the CI build to pass and we're set 👍

PR #11, opened 2023-04-10
feat: ZNSPriceOracle

JamesEarle:
The ZNSPriceOracle contract allows ZNS to get the price of a domain based on the length of that domain. Two different base fees exist, one for root domains and one for subdomains. We call this the `basePrice`

Some amount of domains that are less than a certain length will all be the same, we call this a `baseLength`

There also exists a price multiplier we can use to modify how steep the price drop is in domain names from the initial base price. By default, this value is `3.9` because it forms a reasonable price decline.

We use a simple formula to calculate a domain's price: `(priceMultiplier * basePrice) / length`

Because no decimal values exist in Solidity, we increase by an order of 10 when storing the value for the multiplier and decrease by the same amount in the call to get a domain's price.

This is a visual of the curve of prices with a default `baseLength` of 3

image

### DIFF
--- a/contracts/IZNSPriceOracle.sol
+++ b/contracts/IZNSPriceOracle.sol
@@ -2,9 +2,11 @@
 pragma solidity ^0.8.18;
 
 interface IZNSPriceOracle {
-  event BasePriceSet(uint price, bool isSubdomain);
-  event PriceMultiplierSet(uint multiplier);
-  event BaseLengthSet(uint8 length);
+  event BasePriceSet(uint256 price, bool isSubdomain);
+  event PriceMultiplierSet(uint256 multiplier);
+  event BaseLengthSet(uint8 length, bool isRootDomain);
+  event BaseLengthsSet(uint8 rootDomainLength, uint8 subdomainLength);
+  event ZNSRegistrarSet(address registrar);
 
   /**
    * @notice Get the price of a given domain name length
@@ -12,9 +14,9 @@ interface IZNSPriceOracle {
    * @param isRootDomain Flag for which base price to use. True for subdomain, false for root.
    */
   function getPrice(
-    uint8 length,
+    string calldata length,
     bool isRootDomain
-  ) external view returns (uint);
+  ) external view returns (uint256);
 
   /**
    * @notice Set the base price for root domains
@@ -23,10 +25,7 @@ interface IZNSPriceOracle {
    * @param basePrice The price to set in $ZERO
    * @param isRootDomain Flag for if the price is to be set for a root or subdomain
    */
-  function setBasePrice(
-    uint basePrice,
-    bool isRootDomain
-  ) external;
+  function setBasePrice(uint256 basePrice, bool isRootDomain) external;
 
   /**
    * @notice In price calculation we use a `multiplier` to adjust how steep the
@@ -35,14 +34,22 @@ interface IZNSPriceOracle {
    *
    * @param multiplier The new price multiplier to set
    */
-  function setPriceMultiplier(uint16 multiplier) external;
+  function setPriceMultiplier(uint256 multiplier) external;
 
   /**
    * @notice Set the value of the domain name length boundary where the default price applies
    * e.g. A value of '5' means all domains <= 5 in length cost the default price
    * @param length Boundary to set
+   * @param isRootDomain Flag for if the price is to be set for a root or subdomain
    */
-  function setBaseLength(uint8 length) external;
+  function setBaseLength(uint8 length, bool isRootDomain) external;
+
+  /**
+   * @notice Set the value of both base lengt variables
+   * @param rootLength The length for root domains
+   * @param subdomainLength The length for subdomains
+   */
+  function setBaseLengths(uint8 rootLength, uint8 subdomainLength) external;
 
   /**
    * @notice Set the ZNSRegistrar for this contract
--- a/contracts/IZNSPriceOracle.sol
+++ b/contracts/IZNSPriceOracle.sol
@@ -35,7 +35,7 @@ interface IZNSPriceOracle {
    *
    * @param multiplier The new price multiplier to set
    */
-  function setPriceMultipler(uint multiplier) external;
+  function setPriceMultiplier(uint16 multiplier) external;
 
   /**
    * @notice Set the value of the domain name length boundary where the default price applies
--- a/contracts/IZNSPriceOracle.sol
+++ b/contracts/IZNSPriceOracle.sol
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+interface IZNSPriceOracle {
+  event BasePriceSet(uint price, bool isSubdomain);
+  event PriceMultiplierSet(uint multiplier);
+}
--- a/contracts/IZNSPriceOracle.sol
+++ b/contracts/IZNSPriceOracle.sol
@@ -4,4 +4,5 @@ pragma solidity ^0.8.18;
 interface IZNSPriceOracle {
   event BasePriceSet(uint price, bool isSubdomain);
   event PriceMultiplierSet(uint multiplier);
+  event BaseLengthSet(uint8 length);
 }
--- a/contracts/IZNSPriceOracle.sol
+++ b/contracts/IZNSPriceOracle.sol
@@ -5,4 +5,54 @@ interface IZNSPriceOracle {
   event BasePriceSet(uint price, bool isSubdomain);
   event PriceMultiplierSet(uint multiplier);
   event BaseLengthSet(uint8 length);
+
+  /**
+   * @notice Get the price of a given domain name length
+   * @param length The length of the name to check
+   * @param isRootDomain Flag for which base price to use. True for subdomain, false for root.
+   */
+  function getPrice(
+    uint8 length,
+    bool isRootDomain
+  ) external view returns (uint);
+
+  /**
+   * @notice Set the base price for root domains
+   * If this value or the `priceMultiplier` value is `0` the price of any domain will also be `0`
+   *
+   * @param basePrice The price to set in $ZERO
+   * @param isRootDomain Flag for if the price is to be set for a root or subdomain
+   */
+  function setBasePrice(
+    uint basePrice,
+    bool isRootDomain
+  ) external;
+
+  /**
+   * @notice In price calculation we use a `multiplier` to adjust how steep the
+   * price curve is after the base price. This allows that value to be changed.
+   * If this value or the `basePrice` is `0` the price of any domain will also be `0`
+   *
+   * @param multiplier The new price multiplier to set
+   */
+  function setPriceMultipler(uint multiplier) external;
+
+  /**
+   * @notice Set the value of the domain name length boundary where the default price applies
+   * e.g. A value of '5' means all domains <= 5 in length cost the default price
+   * @param length Boundary to set
+   */
+  function setBaseLength(uint8 length) external;
+
+  /**
+   * @notice Set the ZNSRegistrar for this contract
+   * @param registrar The registrar to set
+   */
+  function setZNSRegistrar(address registrar) external;
+
+  /**
+   * @notice Return true if a user is authorized, otherwise false
+   * @param user The user to check
+   */
+  function isAuthorized(address user) external view returns (bool);
 }
--- a/contracts/ZNSPriceOracle.sol
+++ b/contracts/ZNSPriceOracle.sol
@@ -174,6 +174,10 @@ contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
    */
   function _setZNSRegistrar(address registrar) internal {
     require(registrar != address(0), "ZNS: Zero address for Registrar");
+
+    // Modify the access control for the new registrar
+    authorized[znsRegistrar] =  false;
     znsRegistrar = registrar;
+    authorized[registrar] = true;
   }
 }
--- a/contracts/ZNSPriceOracle.sol
+++ b/contracts/ZNSPriceOracle.sol
@@ -9,28 +9,30 @@ contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
   /**
    * @notice Base price for root domains
    */
-  uint public rootDomainBasePrice;
+  uint256 public rootDomainBasePrice;
 
   /**
    * @notice Base price for subdomains
    */
-  uint public subdomainBasePrice;
+  uint256 public subdomainBasePrice;
 
   /**
    * @notice The price multiplier used in calculation for a given domain name's length
    * We store this value with two decimals of precision for division later in calculation
    * This means if we use a multiplier of 3.9, it is stored as 390
-   * Note that 3.9 is recommended but is an arbitrary choice to use as a multiplier. We use 
+   * Note that 3.9 is recommended but is an arbitrary choice to use as a multiplier. We use
    * it here because it creates a reasonable decline in pricing visually when graphed.
    */
-  uint16 public priceMultiplier;
+  uint256 public priceMultiplier;
 
   /**
    * @notice The base domain name length is used in pricing to identify domains
    * that should recieve the default base price for that type of domain or not.
    * Domains that are `<= baseLength` will receive the default base price.
    */
-  uint8 public baseLength;
+  uint8 public rootDomainBaseLength;
+
+  uint8 public subdomainBaseLength;
 
   /**
    * @notice The address of the ZNS Registrar we are using
@@ -52,39 +54,46 @@ contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
   }
 
   function initialize(
-    uint rootDomainBasePrice_,
-    uint subdomainBasePrice_,
-    uint16 priceMultiplier_,
-    uint8 baseLength_,
+    uint256 rootDomainBasePrice_,
+    uint256 subdomainBasePrice_,
+    uint256 priceMultiplier_,
+    uint8 rootDomainBaseLength_,
+    uint8 subdomainBaseLength_,
     address znsRegistrar_
   ) public initializer {
     rootDomainBasePrice = rootDomainBasePrice_;
     subdomainBasePrice = subdomainBasePrice_;
     priceMultiplier = priceMultiplier_;
-    baseLength = baseLength_;
+    rootDomainBaseLength = rootDomainBaseLength_;
+    subdomainBaseLength = subdomainBaseLength_;
 
     _setZNSRegistrar(znsRegistrar_);
-    
+
     authorized[msg.sender] = true;
     authorized[znsRegistrar_] = true;
   }
 
   /**
    * @notice Get the price of a given domain name length
-   * @param length The length of the name to check
+   * @param name The name of the domain to check
    * @param isRootDomain Flag for which base price to use. True for root, false for subdomains
    */
   function getPrice(
-    uint8 length,
+    string calldata name,
     bool isRootDomain
-  ) external view returns (uint) {
+  ) external view returns (uint256) {
+    // TODO Getting string length this way may be misleading
+    // when considering unicode encoded characters. Find a
+    // better solution for this
+    uint8 length = uint8(bytes(name).length);
+
     // No pricing is set for 0 length domains
     if (length == 0) return 0;
 
     if (isRootDomain) {
-      return _getPrice(length, rootDomainBasePrice);
+      return _getPrice(length, rootDomainBaseLength, rootDomainBasePrice);
     } else {
-      return _getPrice(length, subdomainBasePrice);
+      return _getPrice(length, subdomainBaseLength, subdomainBasePrice);
     }
   }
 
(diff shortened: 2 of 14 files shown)


contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    emit PriceMultiplierSet(multiplier);
+  }
+
+  /**
+   * @notice Set the value of the domain name length boundary where the default price applies
+   * e.g. A value of '5' means all domains <= 5 in length cost the default price
+   * @param length Boundary to set
+   */
+  function setBaseLength(uint8 length) external onlyAuthorized {
```
RyanCross: what's the context of this requirement? What happens if a root domain's name >= then 5? (5 character limit on domain name seems pretty restrictive)

EDIT: I think the meat of what I'm trying to figure out here is: will the root domain always be "zero" in this new architecture?

Whytecrowe: I think we've confused outselves with the terms Neo uses. He calls 2LDs root domains, which means, it's the first level domain after `0://`, which is what we used for naming here (let me know if it's confusing and we should change it).
in reality, of course, they are not roots in the conventional sense. they are 2LDs (e.g. `0://wilder`).
There's not limit for the length other then Solidity's limit of 255, I think.

Not sure what you mean by "always be "zero"".
The base price is set to all domains lower then "base length", meaning that price after string length of 5 chars will be dropping asymptotically, before 5 chars the price is fixed and maximal.

Did I understand your question right? :)

contracts/IZNSPriceOracle.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+interface IZNSPriceOracle {
+  event BasePriceSet(uint price, bool isSubdomain);
+  event PriceMultiplierSet(uint multiplier);
+  event BaseLengthSet(uint8 length);
```
Whytecrowe: please, add all externally called functions to this interface to make it an actual interface

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * @notice Base price for subdomains
+   */
+  uint public subdomainBasePrice;
+
+  /**
+   * @notice The price multiplier used in calculation for a given domain name's length
+   * @dev 3.9 is an arbitrary choice to use as a multiplier. I used it because it created
+   * a reasonable decline in pricing visually when graphed.
```
Whytecrowe: change this for a TODO comment if it's personal thoughts that need to be addressed later. natspec should be facing the user or a developer that builds on top of the current module.
I like the comment, just don't want us to miss it :)

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+   */
+  // ufixed public priceMultiplier = 3.9;
+  uint public priceMultiplier = 39;
```
Whytecrowe: this shouldn't have a value in the code. should only be a settable var, that we initialize at construction time.
same for the one below (`baseLength`). should be set in constructor/init

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * @notice The address of the ZNS Registrar we are using
+   */
+  address public znsRegistrar;
+
+  /**
+   * @notice Track authorized users or contracts
+   */
+  mapping(address => bool) authorized;
```
Whytecrowe: please add a TODO comment with words "access control" in it. there will be a larger task of setting proper access control for the whole system. this should be a part of it. TODO comments will help mark the spots where AC needs to be evaluated.

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+   */
+  modifier onlyAuthorized() {
+    require(authorized[msg.sender], "ZNS: Not authorized");
+    _;
+  }
+
+  function initialize(
+    uint rootDomainBasePrice_,
+    uint subdomainBasePrice_,
+    address znsRegistrar_
```
Whytecrowe, line 62: add `baseLength` and `priceMultiplier`

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+    subdomainBasePrice = subdomainBasePrice_;
+    znsRegistrar = znsRegistrar_;
+    authorized[msg.sender] = true;
+    authorized[znsRegistrar_] = true;
+  }
+
+  /**
+   * @notice Get the price of a given domain name length
+   * @param length The length of the name to check
+   * @param isRootDomain Flag for which base price to use. True for subdomain, false for root.
```
Whytecrowe: wrong comment. should be "true" for root domains

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+  ) internal view returns (uint) {
+    if (length <= baseLength) return basePrice;
+
+    // TODO truncate to everything after the decimal, we don't want fractional prices
+    // Should this be here vs. in the dApp?
+
+    // This creates an asymptotic curve that decreases in pricing based on domain name length
+    // Because there are no decimals in ETH we set the muliplier as an order of magnitutde 
+    // higher than it is meant to be, so we divide by 10 to reverse that action.
+    return (priceMultiplier * basePrice) / length / 10;
```
Whytecrowe: do we have any edges for this formula? any values for which this formula breaks or we get non integer values? should we add some kind of check for these values?

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  function initialize(
+    uint rootDomainBasePrice_,
+    uint subdomainBasePrice_,
+    address znsRegistrar_
+  ) public initializer {
+    rootDomainBasePrice = rootDomainBasePrice_;
+    subdomainBasePrice = subdomainBasePrice_;
+    znsRegistrar = znsRegistrar_;
```
Whytecrowe: check zero address before setting

test/ZNSPriceOracle.test.ts
```diff
@@ -0,0 +1,235 @@
...
+      priceSubdomain = await contract.getPrice(domainB.length, false);
+      expect(priceSubdomain).to.eq(subdomainPrice);
+    });
+
+    it("Returns the expected price for a domain greater than the base length", async () => {
+      const domain = "wilder";
+      const defaultRootPrice = await contract.rootDomainBasePrice();
+      const defaultMultiplier = await contract.priceMultiplier();
+
+      const expectedPrice = (defaultRootPrice.mul(defaultMultiplier)).div(domain.length).div(10);
```
Whytecrowe: DRY
this can be extracted into a function and reused across all tests instead of copying. we will need this function later for other things and tests.

test/ZNSPriceOracle.test.ts
```diff
@@ -0,0 +1,235 @@
...
+      const price = await contract.getPrice(domain.length, true);
+
+      expect(price).to.eq(expectedPrice);
+    });
+  });
+
+  describe("setBasePrice", () => {
+    it("Allows an authorized user to set the base price", async () => {
+      const newBasePrice = hre.ethers.utils.parseEther("0.7");
```
Whytecrowe: just a tiny nit:
if we destructure needed methods in imports, we can just type `par...` and it will autocomplete to the method we need.
this is instead of typing the whole access chain here `hre.ethers.utils...` or looking for it's usages to copy and paste. time-saving, but also a personal preference...

test/ZNSPriceOracle.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    });
+
+    it("Disallows an unauthorized user to set the base price", async () => {
+      const newMultiplier = BigNumber.from("25");
+
+      const tx = contract.connect(user).setPriceMultipler(newMultiplier);
+      expect(tx).to.be.revertedWith("ZNS: Not allowed")
+    });
+
+    it("Allows setting the multiplier to zero", async () => {
```
Whytecrowe: Just writing thoughts here...
Setting multiplier to 0 will make all domains <= `baseLength` to be priced with `basePrice`, but everything after that will be zero.
We can use this to change pricing models by e.g. making a fixed price for all domains under a certain limit and all longer domains be free.

test/ZNSPriceOracle.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    it("Allows an authorized user to set the length boundary", async () => {
+      const newBoundary = 5;
+
+      await contract.connect(deployer).setBaseLength(newBoundary);
+
+      const updatedBoundary = await contract.baseLength();
+      expect(updatedBoundary).to.eq(newBoundary);
+    });
+
+    it("Disallows an unauthorized user to set the base price", async () => {
```
Whytecrowe: "length"

test/ZNSPriceOracle.test.ts
```diff
@@ -0,0 +1,235 @@
...
+      const newBoundary = 5;
+
+      const tx = contract.connect(user).setBaseLength(newBoundary);
+      expect(tx).to.be.revertedWith("ZNS: Not allowed")
+    });
+
+    it("Allows setting the boundary to zero", async () => {
```
Whytecrowe: this will make all prices 0. do we want there to be another way of doing this, besides changing the `basePrice`?
Maybe not to avoid mistakes?..

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,156 @@
...
+   * @notice The base domain name length is used in pricing to identify domains
+   * that should recieve the default base price for that type of domain or not.
+   * Domains that are `<= baseLength` will receive the default base price.
+   */
+  uint8 public baseLength = 3;
+
+  /**
+   * @notice The address of the ZNS Registrar we are using
+   */
+  address public znsRegistrar;
```
Whytecrowe, line 40: missing a setter function for this

Whytecrowe (changes requested):
Nice clean pretty code! :)
Some minor comments to review.
Please, make sure you run `eslint` with `--fix` flag so it does a bit of reformatting.
Also, make sure you connect eslint locally in your IDE so it highlights errors as you type and also add `eslint --fix` to `actions on save`. Prettier can be turned off for now. We may add it later if needed.

Update this branch from `development`. A new PR has been merged.

contracts/ZNSPriceOracle.sol
JamesEarle: I like it there too, and agree it shouldn't be natspec I guess in that perspective, but also leaving it as TODO makes it feel like an action item that needs resolution, when it's really just explanatory. What do you think? Probably no tag is the best choice IMO

contracts/ZNSPriceOracle.sol
JamesEarle: An effort to reduce the initializer params, they still have setters, but I can update them

contracts/ZNSPriceOracle.sol
JamesEarle: I had the opposite logic at first and reversed it, forgot to also update the comment 👍

test/ZNSPriceOracle.test.ts
JamesEarle: Correct yeah, we could extend the `baseLength` value so all domains that are 5 or 7 characters cost `basePrice` and the rest are free

test/ZNSPriceOracle.test.ts
JamesEarle: If either of the `basePrice` or `priceMultiplier` values are 0, then all domains are free that are above the specified length. If the length is 0 as well, then all domains are free, but if the length is 0 and the prior is not true (i.e. neither `basePrce` or `priceMultiplier` are 0) then we just perform the formula for their price based on the length, we don't return the base price

So it will only make all prices 0 in the case that at least one of `basePrce` or `priceMultiplier` is also 0

contracts/ZNSPriceOracle.sol
JamesEarle: @RyanCross A top level domain like `0://wilder` will always have a `parentId` of `0x0...`, so the "root" will always be zero

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,179 @@
...
+   * @notice Get the price of a given domain name length
+   * @param length The length of the name to check
+   * @param isRootDomain Flag for which base price to use. True for root, false for subdomains
+   */
+  function getPrice(
+    uint8 length,
+    bool isRootDomain
+  ) external view returns (uint) {
+    // No pricing is set for 0 length domains
+    if (length == 0) return 0;
```
beardhero, line 91: Is there a danger here of some external contract or app thinking that a length 0 domain is free and creating it?

I know we're preventing 0 from being set in other contacts, but if some future contract didn't then this might get missed?

Could just revert here instead

contracts/ZNSPriceOracle.sol
JamesEarle: @Whytecrowe If `priceMultiplier` or `basePrice` are 0, then this calculation returns 0. This calculation only gets performed if a domain name is longer than the `baseLength` value currently set. 

If `baseLength` is 0, then this function is always executed. In the case that `baseLength` and one or both of `priceMultplier` or `basePrice` are zero, this will return 0.

From our discussions, this may end up being a useful way to play with the pricing dynamic.

The only way this truly "breaks" is if `length` is zero, which is checked in the above external `getPrice` function

contracts/ZNSPriceOracle.sol
JamesEarle: Good question, I'm not sure. I didn't want to revert here as this is just a `view` function. I interpret this as more "there is no price for a 0 length domain" not "0 length domains are free".

I imagine when we have the Registrar call `registerDomain` or something similar, it would `require(length != 0)` there, before calling this, and it makes more sense to revert there, IMO

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * @notice Base price for root domains
+   */
+  uint public rootDomainBasePrice;
+
+  /**
+   * @notice Base price for subdomains
+   */
+  uint public subdomainBasePrice;
+
```
beardhero, line 18: So it's seems that with this contract, the intention is to deploy a new price oracle for every registrar. 

So what is this distinction between root and sub domains then?

The registrar, which should already be deployed and associated to a domain, should not need to set any prices for its root domains that are already minted.

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,179 @@
...
+  ) public initializer {
+    rootDomainBasePrice = rootDomainBasePrice_;
+    subdomainBasePrice = subdomainBasePrice_;
+    priceMultiplier = priceMultiplier_;
+    baseLength = baseLength_;
+
+    _setZNSRegistrar(znsRegistrar_);
+    
+    authorized[msg.sender] = true;
+    authorized[znsRegistrar_] = true;
```
beardhero, line 73: So with this access pattern, a new price oracle has to be deployed for every new registrar?

Then shouldn't this oracle contract just get inherited by the registrars. They will at least have to import them to get setter access.

This means that the oracle needs the registrar to initialize and the registrar needs the oracle to initialize, which is a problem right?

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,179 @@
...
+    uint basePrice
+  ) internal view returns (uint) {
+    if (length <= baseLength) return basePrice;
+
+    // TODO truncate to everything after the decimal, we don't want fractional prices
+    // Should this be here vs. in the dApp?
+
+    // This creates an asymptotic curve that decreases in pricing based on domain name length
+    // Because there are no decimals in ETH we set the muliplier as an order of magnitutde 
+    // higher than it is meant to be, so we divide by 10 to reverse that action.
```
beardhero: Instead of magic number 10, set a constant priceDivisor and consider automatically setting it to a factor of the priceMultiplier when that's set.

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,179 @@
...
+
+  /**
+   * @notice In price calculation we use a `multiplier` to adjust how steep the
+   * price curve is after the base price. This allows that value to be changed.
+   * If this value or the `basePrice` is `0` the price of any domain will also be `0`
+   *
+   * @param multiplier The new price multiplier to set
+   */
+  function setPriceMultipler(uint multiplier) external onlyAuthorized {
+    priceMultiplier = multiplier;
```
beardhero, line 136: This can break the price calc if set too low.

Prevent setting 0 and <baseLength*priceDivisor

contracts/ZNSPriceOracle.sol
```diff
@@ -0,0 +1,179 @@
...
+   */
+  function setBasePrice(
+    uint basePrice,
+    bool isRootDomain
+  ) external onlyAuthorized {
+    if (isRootDomain) {
+      rootDomainBasePrice = basePrice;
+    } else {
+      subdomainBasePrice = basePrice;
+    }
```
beardhero, line 115: Can break price calc if set too low

Prevent setting 0 and something like <baseLength*priceDivisor

contracts/ZNSPriceOracle.sol
JamesEarle: @beardhero We always implicitly have to multiply by 10 to store the value we actually want, because we cant use fractional values like float or double, and so we have to always divide by 10 as well so this value is intentionally not variable.

contracts/ZNSPriceOracle.sol
JamesEarle: Resolving based on discussion in standup, the price oracle will be used only by our registrar

contracts/ZNSPriceOracle.sol
JamesEarle: Per our conversation in standup, the price oracle will only be used by our main registrar

contracts/ZNSPriceOracle.sol
JamesEarle: From conversation with Kirill, the calc is meant to be able to return 0 in case we want to play with the way pricing works in any way like offering free domains for early adopters etc.

contracts/ZNSPriceOracle.sol
Whytecrowe: Even if Registrar does not check for 0, how would a zero length name even exist in the system? How would they access it?
So, they do create a free no name domain. what's next? can they actually use it?

contracts/ZNSPriceOracle.sol
Whytecrowe: Not every Registrar has to have an Oracle. If someone wants to do it this complicated way - they have the right, but they are not obligated. they can handle the prices in whichever way they want.

Regarding initialization, I don't see it as a problem, we can set this as a part of our deployment process to initialize for one contract and then just use a setter for another one. Am I missing something here?

contracts/ZNSPriceOracle.sol
Whytecrowe: I don't see why. the divisor here is to eliminate potential problems with decimals, right @JamesEarle ?
Why is it 10 and not a higher number like 1000? is one decimal enough for us to eliminate fractions? it doesn't seem like it's the case

contracts/ZNSPriceOracle.sol
Whytecrowe: maybe not 0, though. setting this to zero will allow us to eliminate prices for all domains after `baseLength`, which might be helpful, but I don't see when and why yet... either way, it's not "breaking" anything per se...
we do NOT want to have a 0, however, if we do set `priceMultiplier` as non-zero. hence my previous comments on what is the range of our actual valid values we can set here?..

contracts/ZNSPriceOracle.sol
JamesEarle: @Whytecrowe This isn't for truncating to avoid fractional values, this is just for storing the multiplier. We *technically* can use any value for as many decimals of precision we want to store. If we want a multiplier to be something more precise, we can use 100,000 instead of 10, we just have to divide by the same number here when we call to getPrice.

I can just make it a variable if we want it to be adjustable, but I only see value in changes that either increase or decrese the amount of precision kept (e.g. 100 => 1000, not 100 => 150)

contracts/ZNSPriceOracle.sol
beardhero: To keep the namehash process consistent, I think the root should be namehash("0://") instead of just 0. 
This makes it clear how we would deploy a new root.

contracts/ZNSPriceOracle.sol
beardhero: I see what you mean that you're just trying to set the effective multiplier to 3.9, but also we have to be aware of the fraction truncation and dust - otherwise you would just set the price multiplier lower and save gas.

In general constants shouldn't be set like this in views because it's not upgrade safe in the case that some other contract tries to save gas by doing this price calculation offchain, even though that may not be as common.

Another reason is just the same as why we store the multiplier, because you might want this to change and so it needs a setter.

Side note: remove unnecessary parentheses

contracts/ZNSPriceOracle.sol
JamesEarle: I don't think there is a threat here, it's the equivalent of owning a domain that is `0x0...` and it doesn't have any effect as you wouldn't be able to use it. Plus, in the Registry we declare the owner of `0x0` to be the deployer of that contract and only they'd be able to modify who owns it, so no one can call to register it even if they could get a price here.

PR #15, opened 2023-04-20
Adding String Length Check Utility

Will-Hyde:
Adding a utility that can check the length of strings accurately, an example has been taken from the ENS implementation as it seems to be very well received and reviewed

### DIFF
--- a/contracts/StringUtils.sol
+++ b/contracts/StringUtils.sol
@@ -2,6 +2,8 @@
 // Source:
 // https://github.com/ensdomains/ens-contracts/blob/master/contracts/ethregistrar/StringUtils.sol
 pragma solidity ^0.8.18;
+
+
 library StringUtils {
     /**
      * @dev Returns the length of a given string
--- a/contracts/StringUtils.sol
+++ b/contracts/StringUtils.sol
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+// Source:
+// https://github.com/ensdomains/ens-contracts/blob/master/contracts/ethregistrar/StringUtils.sol
+pragma solidity ^0.8.18;
+library StringUtils {
+    /**
+     * @dev Returns the length of a given string
+     *
+     * @param s The string to measure the length of
+     * @return The length of the input string
+     */
+    function strlen(string memory s) internal pure returns (uint256) {
+        uint256 len;
+        uint256 i = 0;
+        uint256 bytelength = bytes(s).length;
+        for (len = 0; i < bytelength; len++) {
+            bytes1 b = bytes(s)[i];
+            if (b < 0x80) {
+                i += 1;
+            } else if (b < 0xE0) {
+                i += 2;
+            } else if (b < 0xF0) {
+                i += 3;
+            } else if (b < 0xF8) {
+                i += 4;
+            } else if (b < 0xFC) {
+                i += 5;
+            } else {
+                i += 6;
+            }
+        }
+        return len;
+    }
+}
--- a/contracts/ZNSPriceOracle.sol
+++ b/contracts/ZNSPriceOracle.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.18;
 import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import { IZNSPriceOracle } from "./IZNSPriceOracle.sol";
-
+import { StringUtils } from "./StringUtils.sol";
 contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
   /**
    * @notice Base price for root domains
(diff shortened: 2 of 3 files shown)


contracts/ZNSPriceOracle.sol
```diff
@@ -82,18 +82,16 @@ contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
     string calldata name,
     bool isRootDomain
   ) external view returns (uint256) {
-    // TODO Getting string length this way may be misleading
-    // when considering unicode encoded characters. Find a
-    // better solution for this
-    uint8 length = uint8(bytes(name).length);
-
+    uint256 length = StringUtils.strlen(name);
+    require(length <= 255, "ZNS: Only names up to 255 characters in length are supported");
```
JamesEarle: A string would be one whole word in storage so 256 bits. It shouldn't be possible to even form the transaction if the string is longer, so this may be redundant because it can never be hit. There is already a `long domain name` test in the oracle here, try increasing the length and seeing if it even creates a transaction for you.

Whytecrowe: good point! was thinking about it as well

contracts/StringUtils.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+// Source:
+// https://github.com/ensdomains/ens-contracts/blob/master/contracts/ethregistrar/StringUtils.sol
+pragma solidity ^0.8.18;
+library StringUtils {
```
Whytecrowe, line 7: let's add 2 empty lines here and put source under the pragma

contracts/ZNSPriceOracle.sol
```diff
@@ -82,18 +82,16 @@ contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
     string calldata name,
     bool isRootDomain
   ) external view returns (uint256) {
-    // TODO Getting string length this way may be misleading
-    // when considering unicode encoded characters. Find a
-    // better solution for this
-    uint8 length = uint8(bytes(name).length);
-
+    uint256 length = StringUtils.strlen(name);
```
Whytecrowe: we should change this by adding this line:
`using StringUtils for string;`    right under the contract name, then go:
`name.strlen()` on this line.

This will be a good base if we need to add any other utils in the future. this will also make a `delegatecall` to the library preserving the context and state.
When we outline it like this we are using it as a library rather then contract, which makes txes cheaper.

contracts/ZNSPriceOracle.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.18;
 import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import { IZNSPriceOracle } from "./IZNSPriceOracle.sol";
-
+import { StringUtils } from "./StringUtils.sol";
 contract ZNSPriceOracle is IZNSPriceOracle, Initializable {
```
Whytecrowe, line 10: always two empty lines before `contract` keyword

Whytecrowe (changes requested):
some minor things in comments

RyanCross (approved):

contracts/ZNSPriceOracle.sol
JamesEarle: I tested this locally and I was wrong, you can form a tx with a string longer than 256 bits, and it was only failing my tests because I was using `uint8` to hold the length. Weird though is that I swore it was actually failing before it even formed the tx when I went above that length even before reducing the size of the variable.

The question now is should we enforce domain name length? if so, we can keep the vars as `uint8` and then keep this require statement as well, as otherwise it causes overflow for the `uint8`, otherwise we should remove it and set the vars to `uint256` instead.

@Whytecrowe @Will-Hyde ?

JamesEarle: I tested this locally and I was wrong, you can form a tx with a string longer than 256 bits, and it was only failing my tests because I was using `uint8` to hold the length. Weird though is that I swore it was actually failing before it even formed the tx when I went above that length even before reducing the size of the variable.

The question now is should we enforce domain name length? if so, we can keep the vars as `uint8` and then keep this require statement as well, as without it causes overflow for the `uint8` and at 257 chars the domain is free, but at 258 the cost is `basePrice` again. Otherwise we should remove the require statement and set the vars to `uint256` instead.

@Whytecrowe @Will-Hyde ?

contracts/StringUtils.sol
Will-Hyde: Added

contracts/ZNSPriceOracle.sol
Will-Hyde: Updated to call name.strlen()

contracts/ZNSPriceOracle.sol
Will-Hyde: Added

PR #17, opened 2023-04-26
Mud 195/domain exemption logic

Will-Hyde:
Validating and writing tests for domain revoking


contracts/token/IZNSDomainToken.sol
```diff
@@ -2,7 +2,11 @@
 pragma solidity ^0.8.18;
 
 interface IZNSDomainToken {
+  event SetAuthorization(address account);
```
RyanCross: index account

RyanCross: this event name is a little too ambiguous for me. What exactly are we authorizing?

contracts/registry/ZNSRegistry.sol
```diff
@@ -104,7 +104,9 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
   function deleteRecord(bytes32 domainNameHash) external {
     // this could call to an internal func _deleteDomainRecord
     // Then when `setDomainRecord` is `0x0` values, we can also delete there
-    require(msg.sender == records[domainNameHash].owner, "ZNSRegistry");
+    
+    //This doesnt work.
+    //require(msg.sender == records[domainNameHash].owner;
```
RyanCross, line 109: Can you add some explanation as to why this doesn't work?  What's stopping any joe schmoe from calling deleteRecord now?

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.18;
 
-import { IZNSEthRegistrar } from "./IZNSEthRegistrar.sol";
-import { IZNSRegistry } from "../registry/IZNSRegistry.sol";
-import { IZNSTreasury } from "./IZNSTreasury.sol";
-import { IZNSDomainToken } from "../token/IZNSDomainToken.sol";
-import { IZNSAddressResolver } from "../resolver/IZNSAddressResolver.sol";
-import { IZNSPriceOracle } from "./IZNSPriceOracle.sol";
+import {IZNSEthRegistrar} from "./IZNSEthRegistrar.sol";
```
Whytecrowe: let's keep the spacing around curly braces how it is on other contracts

contracts/registry/ZNSRegistry.sol
Whytecrowe: this is a perfect case for the future Access Control task. we need to make sure only Registrar can call this. right now it is not protected.
and it doesn't work because the call comes from ZNSRegistrar and it is not the owner of the domain

contracts/token/IZNSDomainToken.sol
Whytecrowe: this is a temporary solution that needs to be reworked when doing access control

contracts/token/ZNSDomainToken.sol
```diff
@@ -10,7 +10,36 @@ import { IZNSDomainToken } from "./IZNSDomainToken.sol";
 contract ZNSDomainToken is ERC721, IZNSDomainToken {
   // TODO: change for proper name !
   // solhint-disable-next-line no-empty-blocks
-  constructor() ERC721("ZNSDomainToken", "ZDT") {}
+  constructor() ERC721("ZNSDomainToken", "ZDT") {
```
Whytecrowe: these values for name and symbol should be variables that we set at construction time and pass from the deployer wallet

contracts/registry/ZNSRegistry.sol
```diff
@@ -21,7 +21,7 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
    * @notice Mapping of `owner` => `operator` => `bool` to show accounts that
    * are or aren't allowed access to domains that `owner` has access to.
    */
-  mapping(address owner => mapping(address operator => bool)) private operators;
+  mapping(address owner => mapping(address operator => bool b) m) private operators;
```
Whytecrowe: what's `b` and what's `m` ? if we do use var names, let's make them clear.
like, a bool can be `allowed` and `m` is not required since we don't really care to name a mapping inside a mapping

contracts/token/IZNSDomainToken.sol
Will-Hyde: indexed

Will-Hyde: Gave it a more clear name, its just meant to authorize a contract to be able to execute methods

contracts/registry/ZNSRegistry.sol
Will-Hyde: Yes thats correct, the ZNSRegistrar otherwise doesnt have access to call this function. Ill expand the comment

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: I think my linter was removing it automatically, fixed now

contracts/token/ZNSDomainToken.sol
Will-Hyde: Now set by constructor from a variable

contracts/registry/ZNSRegistry.sol
Will-Hyde: Theyre just generic names because otherwise I get a lint issue, I can provide a more substantial name

RyanCross (approved):
Looks good after merge conflicts resolved.

PR #18, opened 2023-05-01
feat: Access Control #1

Whytecrowe:
Roles and AccessControlled contracts should be inherited by every contract to have a properly implemented access control, so they all have access to the AccessManager contract, it's functionality and all the roles outlined


RyanCross:
@Whytecrowe I wanted to summarize some of conversations https://github.com/zer0-os/zNS/pull/21/files since this seems the more appropriate spot. Please correct any misunderstanding.

My general understanding is:

1. Owner/Operator roles are our access control for users and relate strictly to **domain management** outside of **creation**. (Modification, setting resolvers)
2. Governor/Admin act similarly as our access control roles for **system management** for both handling upgrades and system wide workflows (Setting prices, initial minting)
3. Admins _can_ be contracts
4. Governors and Admins have no access to transact on domain modification operations

Some thoughts and questions:

- In MVP Operator should have access to all domain management related functions besides the management of operators and the change of ownership
- UUPS Proxies might slot well into our plan for upgrade management as they exist on the implementation and we can define new upgrade logic (with finer tuned Role-based control) later down the line if needed

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -179,21 +187,21 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
         return domainHash;
     }
 
-    function revokeDomain(bytes32 domainHash) external onlyOwner(domainHash) {
+    function revokeDomain(bytes32 domainHash)
+    external
+    override
+    onlyNameOwner(domainHash)
+    onlyTokenOwner(domainHash)
```
JamesEarle, line 104: Do we really want this to be available for both the name owner and the token owner?

If the division of ownership is to allow giving use of the domain to another contract/user while the actual "domain" still exists in someone's wallet through them still having the token, then they're the actual owner of the domain and only they should have the right to revoke that domain.

contracts/distribution/ZNSTreasury.sol
```diff
@@ -89,32 +72,40 @@ contract ZNSTreasury is IZNSTreasury {
     function unstakeForDomain(
         bytes32 domainHash,
         address owner
-    ) external onlyRegistrar {
+    ) external override onlyRole(REGISTRAR_ROLE) {
```
JamesEarle, line 75: Most of the AC feels good to me except the `REGISTRAR_ROLE`. If there is only ever one registrar a system cares about, why make this a role? A role to me implies it could apply to several acters and maybe this is to allow future proofing, like we might have several registrars in the future but this is the only "role" that applies to contract(s), not users. So in the worst case, say an account was compromised somehow, the role for the registrar could be given to another malicious contract that is then allowed to call functions like this.

Another note is that we accept an `owner` parameter here instead of actually finding the name owner through the ZNSRegistry, and this would make the transfer below send to anyone that the caller specifies. I don't think this is what is intended

PR #19, opened 2023-05-02
Replace test expected values with ens-namehash lib

RyanCross:
Ensure our on chain hashing results in predictable output similar to ens by using the ens-namehash-lib 

As far as string validation goes: the namehash lib only provides the following guarantees: case insensitivity, normalization of characters to UTF-8, and disallowing non UTF-8 characters. Ens namehashing provides no guarantees of protections against things like homoglyph attacks or sanity checking against slurs and curses, we would need to do this ourselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "typescript": "^5.0.2"
   },
   "dependencies": {
-    "@ensdomains/ensjs": "^2.1.0",
+    "@ensdomains/ensjs": "2.1.0",
     "dotenv": "16.0.3"
   },
   "resolutions": {
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "typescript": "^5.0.2"
   },
   "dependencies": {
+    "@ensdomains/ensjs": "^2.1.0",
     "dotenv": "16.0.3"
   },
   "resolutions": {
--- a/package.json
+++ b/package.json
@@ -49,10 +49,10 @@
     "solidity-coverage": "^0.8.2",
     "ts-node": "10.9.1",
     "typechain": "8.1.0",
-    "typescript": "^5.0.2"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -52,9 +52,10 @@
     "typescript": "^5.0.2"
   },
   "dependencies": {
+    "@ensdomains/ensjs": "2.1.0",
```
Whytecrowe: should probably a devDep, since we're only using it in tests?
or would you like to see it as part of this package when we import ZNS elsewhere?

Whytecrowe: I guess, we can keep it here, since we don't have another repo for now and we can just reuse this with the helper function. disregard my previous message

test/helpers/index.ts
```diff
@@ -28,9 +27,23 @@ export const getPrice = async (
...
\ No newline at end of file
+};
+
+/**
+ * The ens lib takes the inverse of our domain name format to
+ * produce the same namehash as the one produced from our
+ * registrar contract, so we need to inverse the input here.
+ *
+ */
+export const reverseInputName = (name : string) => {
```
Whytecrowe, line 48: what if instead of just reversing the unput, we make this an entry point to all hashing?
we call this `getDomainHash()` and we would reverse before we call `namehash()` in the same helper.
this way we we will just have one call for any test, and this helper will use ensjs under the hood.
what do you think?

package.json
RyanCross: Actually I agree it should be a devdep. I just forgot to push that change. :)

test/helpers/index.ts
RyanCross: Good call. Added hashDomainLabel and hashDomainName

Whytecrowe (approved):

PR #20, opened 2023-05-04
Feat/mud 222 reclaim logic

Will-Hyde:
Adding logic and tests for reclaim functionality


contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -202,6 +202,23 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
   }
 
+  //Todo: Access Control
+  function reclaimDomain(bytes32 domainHash) external {
+    require(
+      znsRegistry.exists(domainHash),
+      "ZNSEthRegistrar: Domain does not exist"
+    );
+    uint256 tokenId = uint256(domainHash);
+    require(znsDomainToken.isOwner(tokenId, msg.sender), "ZNSEthRegistrar: Not owner of domain");
```
Whytecrowe: imo, this should be the first check.
Also, in the message, I think, more clear would be to say "Not a token owner" rather than "domain". the caller is technically only trying to be the full domain owner at this point

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -202,6 +202,23 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
+    address owner = znsRegistry.getDomainRecord(domainHash).owner;
+    require(owner != msg.sender, "ZNSEthRegistrar: Caller already owner of domain registry");
```
Whytecrowe: "Domain is already owned by the caller"

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -202,6 +202,23 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
+
+    znsAddressResolver.setAddress(domainHash, msg.sender);
```
Whytecrowe: this is not the best assumption here. a person can own a domain, but assign it to a contract or something else.
this should either not be included at all or we need to provide another parameter to this function to specifically set resolver to a certain address. in this case there should also be an option to provide zero address, if zero address is provided, then we do not set anything on the resolver. they can set it later at their own time.

look here for reference of what I mean - https://github.com/zer0-os/zNS/blob/feat/ZNSRegistrar/contracts/distribution/ZNSEthRegistrar.sol#L225-L246

contracts/token/ZNSDomainToken.sol
```diff
@@ -40,6 +40,18 @@ contract ZNSDomainToken is ERC721, IZNSDomainToken {
     emit SetAccessAuthorization(account);
   }
 
+  /**
+   * @notice Checks if provided address is an owner of the provided domain token
+   * @param tokenId The identifying tokenId of a domain
+   * @param candidate The address for which we are checking access
+   */
+  function isOwner(
```
Whytecrowe: is this function necessary?
can't we just do:
`require(znsDomainToken.ownerOf(tokenId) == msg.sender)`  in the contract where we need to check?

test/ZNSEthRegistrar.test.ts
```diff
@@ -367,9 +367,137 @@ describe("ZNSEthRegistrar", () => {
...
+      // Transfer the domain token
+      await zns.domainToken.connect(deployer).transferFrom(deployer.address, user.address, tokenId);
+
+      // Reclaim the Domain
+      await zns.registrar.connect(user).reclaimDomain(domainHash);
+      // Verify domain token is still owned
+      const owner  = await zns.domainToken.connect(user).ownerOf(tokenId);
+      expect(owner).to.equal(user.address);
+
+      // Verify domain is owned in registrar
```
Whytecrowe: "in registry"

test/ZNSEthRegistrar.test.ts
```diff
@@ -367,9 +367,137 @@ describe("ZNSEthRegistrar", () => {
...
+      expect(addressResolverAddress).to.equal(user.address);
+    });
+
+    it("Reclaiming domain token emits DomainReclaimed event", async () => {
+      const topLevelTx = await defaultRootRegistration(deployer, zns, defaultDomain);
+      const domainHash = await getDomainHash(topLevelTx);
+      const tokenId = await getTokenId(topLevelTx);
+
+      // Transfer the domain token
+      await zns.domainToken.connect(deployer).transferFrom(deployer.address, user.address, tokenId);
```
Whytecrowe, line 449: this doesn't have to be `transferFrom()` can just be `transfer()` when the owner of the token transfers himself. this way we do not need an extra `approve()` transaction.
`transferFrom()` is used when somebody else transfers your token

test/ZNSEthRegistrar.test.ts
```diff
@@ -367,9 +367,137 @@ describe("ZNSEthRegistrar", () => {
...
+      const tx = zns.registrar.connect(deployer).reclaimDomain(domainHash);
+
+      // Verify Domain is not reclaimed
+      await expect(tx).to.be.revertedWith("ZNSEthRegistrar: Caller already owner of domain registry");
+
+      // Verify domain is not owned in registrar
+      const registryOwner = await zns.registry.connect(user).getDomainOwner(domainHash);
+      expect(registryOwner).to.equal(deployer.address);
+    });
+  });
```
Whytecrowe, line 558: need another test where domain is tranferred, reclaimed and the revoked to make sure the new owner can unstake. check balances in this one to make sure all transfers happened properly

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: Ok, I moved this check  up and changed the require statement

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: Modified

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: I guess im confused here, are you saying I should or shouldnt set the address on the resolver? Or are you saying I shouldnt unless I can verify its not a contract address?

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: Actually, on second thought:
If I move this statement up, then the error message that bubbles up is "ERC721: Invalid Token ID" which means that I should either keep it where it is or remove the earlier line entirely (Otherwise it serves no purpose because it will never trigger)

contracts/token/ZNSDomainToken.sol
Will-Hyde: ownerOf is not exposed by the DomainToken Interface and is instead on the ERC721 interface.

So if I wanted to call that here, I would need to make DomainToken implement some function that calls ownerOf on the token in some way, I just thought this was a simple way to expose what we needed

test/ZNSEthRegistrar.test.ts
Will-Hyde: fixed

test/ZNSEthRegistrar.test.ts
Will-Hyde: As far as I can tell the Transfer method has 3 optional arguments, (from to and tokenID). Which are all the same arguments, also, it doesnt appear that transfer is exposed publicly by ERC721 and is instead an internal method, I could write a wrapper for it and expose it though. Is that what I should do?

test/ZNSEthRegistrar.test.ts
Will-Hyde: Added

contracts/distribution/ZNSEthRegistrar.sol
Whytecrowe: I guess, I'm saying to think on this a bit more.

Since we are planning multiple resolvers of different types, we can't future proof the function arguments type-wise. We either taking an address to the domain content and assume they will bind it to an address OR we not take any args related to the domain content, leaving the user to set it themselves at a different time on Resolver directly. Another option here is to offer a default way of setting an address only, if a user wishes to do so, but there should be an option for him to not provide it. In case he wants to tie it to the bytes32 data or a string or anything else, and not address. When he provides 0, we assume he wants to do that himself at another time and do not set anything.

Does it make sense?

contracts/token/ZNSDomainToken.sol
Whytecrowe: well, this is a bigger problem then, if I understand this correctly. we want our DomainToken to be fully ERC-721 compatible. if we can't access basic 721 functions, this is not an NFT token. we should inherit the ERC-721 interface by ZNSDomainToken interface and then this interface is inherited by ZNSDomainToken impl contract. this way we will catch all the missing functions. if a certain function is not exposed in the ERC-721 base contract, we need to expose it in the ZNSDomainToken.
I see that IERC721 is NOT inherited by the IZNSDomainToken. which is not good. we should definitely do that and then expose or add functions we are missing

test/ZNSEthRegistrar.test.ts
Whytecrowe: my bad, incorrectly assumed this was erc20 for some unknown reason.
you are right. let's use `safeTransferFrom()` then. look into `onERC721Received` specifications.
this might not be necessary, but we need to be sure. this is partially to cover some old contracts that did incorrect returns  and some other things, like adding some confidence that a receiving contract can handle the transferred tokens.
Let me know what you think on this. Should we implement it or not?

Whytecrowe (changes requested):
After all the changes are done, please, update the branch from `development`.
Make sure that Registrar PR has been merged at that point before you update.

contracts/distribution/ZNSEthRegistrar.sol
Whytecrowe: I guess this error is fired by the `ownerOf()` when we are passing an ID of token that doesn't exist. which is totally fine, right?
we are checking an owner here for tokens that do exist. check if you have a test for that specifically.
if we have a token and somebody other than it's owner is trying to access this function, it should fail in that require.
Am I assuming correctly here?

test/ZNSEthRegistrar.test.ts
Whytecrowe: we technically do not transfer this token to any of our contracts. so if we do implement this, whatever contract receives it at any point, has to have `onERC721Received` implemented. so we may not need it, however safeTransfer is usually a better approach.

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: Yes thats right, thats what its checking. And yes I have a test case that covers if someone tries to reclaim the domain without owning the token. I think this code should stay the same

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: I dont think it makes sense to me, im not sure what this has to do with the current story. Are you asking me to make a change here or no? If were going to change how resolvers work, should I just leave this in and have us update it later?

contracts/token/ZNSDomainToken.sol
Will-Hyde: Ok I can have the DomainToken interface inherit from IERC-721 and that should  expose it

test/ZNSEthRegistrar.test.ts
Will-Hyde: image

Yes if the receiver address is a contract, it needs to inherit onERC721Received

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -190,6 +190,22 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
+    require(
+      znsRegistry.exists(domainHash),
+      "ZNSEthRegistrar: Domain does not exist"
+    );
+    uint256 tokenId = uint256(domainHash);
+    require(znsDomainToken.ownerOf(tokenId) == msg.sender, "ZNSEthRegistrar: Not owner of Token");
+    address owner = znsRegistry.getDomainRecord(domainHash).owner;
+    require(owner != msg.sender, "ZNSEthRegistrar: Domain is already owned by the caller");
+
+    znsRegistry.setSubdomainOwner(domainHash, domainHash, msg.sender);
```
Whytecrowe: why are parent and child the same domain here? this is an interesting case...
parentHash is used in `setSubdomainOwner` to check the access. here you are tricking it to check the access of the current domain. this this raise questions for you when you wrote it? is it fine that Registry access checks can be circumvented?

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: Removing the call to the addressResolver

test/ZNSEthRegistrar.test.ts
```diff
@@ -409,9 +409,166 @@ describe("ZNSEthRegistrar", () => {
...
+      const topLevelTx = await defaultRootRegistration(deployer, zns, defaultDomain);
+      const domainHash = await getDomainHashFromEvent(topLevelTx);
+      const tokenId = await getTokenIdFromEvent(topLevelTx);
+      const staked = await zns.treasury.stakedForDomain(domainHash);
+
+      // Transfer the domain token
+      await zns.domainToken.connect(deployer).transferFrom(deployer.address, user.address, tokenId);
+
+      // Reclaim the Domain
+      await zns.registrar.connect(user).reclaimDomain(domainHash);
```
Whytecrowe, line 428: before doing this, let's check who the original owner is of this domain in Registry to validate.

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: The "parent" in this case (in the setSubdomainOwner function) is only being used to determine who owns the domain. In this case, im using the child for both because I only care if the message sender is the owner of the domain in question

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -190,6 +190,22 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
     // TODO: what are we missing here?
   }
 
+  //TODO: Access Control
+  function reclaimDomain(bytes32 domainHash) external {
```
Whytecrowe, line 194: should we take `tokenId` as an argument here? a person who received the token might not know what the domain is or it's hash, but he should know `tokenId` of the token he owns. what do you think?

Will-Hyde: I think I made it domainHash, because revokeDomain also takes domainHash (and also all the other functions in this contract)

Personally, I would want to keep the methods similar. It seems the rest are all using domainHash right now

test/ZNSEthRegistrar.test.ts
Will-Hyde: Ok, I have added this check

contracts/distribution/ZNSEthRegistrar.sol
Whytecrowe: sorry, edited the comment above

Will-Hyde: Yes, I think its a security issue. Because this method is exposed externally, and that means any caller can call it to set the owner of any other domain (by passing it their own domain as parent to prove ownership).

I also think this method does not relate to my story, and this function should be addressed in the story that sets appropriate access control

Whytecrowe: I think the main issue is our Registry functions... we need a better access checks there for owners.
The only reason this call passes is because in tests Registrar is set to be the operator of all `user` domains.
in reality, it will not be, so it will always fail.

Will-Hyde: Ill add a todo, to make sure this gets addressed when AccessControl is implemented

Whytecrowe: the parent hash in this case should be ROOT_HASH which Registrar actually owns/operates by default. the current call will not work in real world, because we can't own or operate on users' domains. we only own and operate the root (0://).

contracts/distribution/ZNSEthRegistrar.sol
Whytecrowe: ok. they are technically interchangeable.

test/ZNSEthRegistrar.test.ts
```diff
@@ -430,7 +592,7 @@ describe("ZNSEthRegistrar", () => {
       expect(exists).to.be.false;
     });
 
-    it("Revokes a SubDomain - Happy Path", async () => {
+    it ("Revokes a SubDomain - Happy Path", async () => {
```
Whytecrowe: why whitespaces here? they are not on any other tests.

contracts/distribution/ZNSEthRegistrar.sol
Will-Hyde: Ok sure, Ill update the parenthash to be the root hash

test/ZNSEthRegistrar.test.ts
Will-Hyde: Not sure, when I merged it added them. I removed them now

PR #21, opened 2023-05-05
Access Control Spike Discussion - DO NOT MERGE

RyanCross:
The investigation results of the Access Control Spike can be found in `access-control.md`.  Every op on the existing contracts has been given a recommended access control security level.  In addition, there is some analysis and recommendations provided at the end. 

The goal of this spike discussion is to finalize enough AC design decisions to complete the zNS MVP (Note this excludes subdomains)

 Please comment your thoughts, my analysis, or suggest alternate approaches to access control.

### DIFF
--- a/access-control.md
+++ b/access-control.md
@@ -2,13 +2,16 @@
 
 Reference Code https://github.com/zer0-os/zNS/pull/18/
 
+**Abstract** 
 Goal: Identify an appropriate level of access control for zNS administration and its consumers for the current suite of functionality. This does not include subdomain minting.  
 
 This analysis assumes a few guiding architecture principles of zNS
 1. Maximize User Ownership:  Operations on entities within zNS are owned (and also the responsibility of) the user. 
 2. Open: Inspection of code yields confidence that these are the building blocks of a decentralized system to associate a domain with any type of content.  This means Zero Administrator access should be limited to operations required to maintain integrity of the system for all users. 
 
-** Operations Analysis: **
+The analysis here presumes the 3 roles of access defined in the reference code - Owner, Operator, and Zero Admin. A Zero admin may be an account or another smart contract. Operators are controlled by the owner and can perform a subset of domain management operations. 
+
+**Operations Analysis**
 Below are access level security recommendations and associated thoughts for zNS MVP operations for today's (5/5/2023) implementation for zNS. 
 * nosec = no access control security needed. 
 * sec = access control recommended, will be followed by a list of roles
@@ -41,24 +44,32 @@ Below are access level security recommendations and associated thoughts for zNS 
     * setZNSRegistrar ()
     * Summary: Zero admin should only have control over initalization/upgradeability based on our principles, everything else can be owner
 
-
-**Analysis**:
-This analysis presumes 3 roles of access - Owner, Operator, and Zero Admin. A Zero admin may be an account or another smart contract. Operators are controlled by the owner and can perform a subset of domain management operations. 
-
-zNSPriceOracle - All price setting operation access can be limited to owners & operators. Zero Admin control can be limited to upgradeability, if there is a need to maintain registrar mappings in the price oracle then it may be appropriate for ZAC to assign those mappings
+**Conclusions and Recommendations**:
+zNSPriceOracle - All price setting operation access can be limited to owners & operators. Zero Admin control can be limited to upgradeability, if there is a need to maintain registrar mappings in the price oracle then it may be appropriate for ZAC to assign those mappings. Monetary operations to guarantee revenue of Zero system (e.g. royalties) should be hardcoded and immutable (or at a minimum separate operations controlled by Zero Admins)
 
 
-zNSAddressResolver - Resolution of zNS name to a system (setting the contract that acts as the resolver) is data that should be the sole propriety of the owner. This is a key relationship of zNS, and answers (in what way does this domain name have meaning? E.g. is it tied to another system of SCs or is it just a name)
-zNSRegistry - Modification of records
+zNSAddressResolver - Resolution of zNS name to a system (setting the contract that acts as the resolver) is data that should be the sole propriety of the owner. This is a key relationship of zNS, and answers (in what way does this domain name have meaning within ethereum? E.g. is it tied to another system of SCs or is it just a name)
 
 
-zNSRegistry - Similar to the address resolver operations that require access control are 
+zNSRegistry - Modification of records and resolvers require such a high degree of trust that a decision needs to be made on what the purview of operators should be. (See aside on operators and owners for recommendations)
+
+
+zNSRegistry - The ownership of the domain record and the resolver(s) of that domain should exclusively be the purview of the owner.  Similarly 
+
+**Aside on Operators and Owners:**
+The analysis of the above these contracts suggests that almost any operation granted to an operator in zNS MVP requires an extreme degree of trust to the point that operators may as well be owners. One possible simplification of access control is to simply allow for multiple owners, and remove the operator role entirely.  
+
+Another possible route is to allow operators to perform any role with the exception of minting, changing domain ownership, and managing operators. Allowing operators full bandwidth to help manage a system, but with owners always have the capability to recover from a breach of trust
 
 Recommendations: Access Control by Zero should be limited to the cases where the user is minting a domain under default contracts that provide some base level of utility in our ecosystem. E.g. Zero Access Control can set a domains resolver to the default if no custom resolver is provided. One these defaults are overriden by the owner. zNS has no access to revoke.
 
-While zNS matures and is in development I can propose two potential paths for admin level access
 
-Option 2:
+**On Zero Administrator Access:**
+
+While the zNS MVP matures and is in development I can propose two potential paths for admin level access
+
+
+Option 1:
  Zero maintains rights to upgradeability over contracts until such a time that Zero Access control decisions are made via DAO or some other multisig. This may provide some assurances to the userbase about our commitment to a decentralized, democratic system, but in reality we are only obscuring centralization that is still there.
 
 Option 2:
--- a/access-control.md
+++ b/access-control.md
@@ -1,0 +1,42 @@
+Operations Overview (subdomain ops are exempted)
+
+High Level Operations
+
+zNSRegistry
+
+zNSDomainToken
+
+zNSAddressResolver (zNS Default Resolver?)
+
+zNS TLD Registrar/SubdomainRegistrar (ZNSEthRegistrar)
+* mint new domains
+* Update name storage of zNSRegistry (TLD only?)
+
+zNSPriceOracle
+
+Reference Code 
+
+Goal: Identify internal operations and assign them sec, no sec
+Identify admin only operations
+Identify operations of specific users
+Identify operations needed to maintain a subset of the system
+
+Internal Operations
+1. zNSEthRegistrar (TLD/Subdomain Registrar)
+2. zNSRegistry
+3. zNSDomainToken
+4. zNSAddressResolver (Default Resolver)
+5. zNSPriceOracle
+
+Cross-Contract Operations
+1. zNSEthRegistrar (TLD/Subdomain Registrar)
+2. zNSRegistry
+3. zNSDomainToken
+4. zNSAddressResolver (Default Resolver)
+5. zNSPriceOracle
+
+
+
+Zero Admin Control (separate contract, called externally, responsible for any zNS system related access, stuff required to maintain integrity)
+
+Are there operations that need to be maintained by the Zero system and restricted only to admins of our system
--- a/access-control.md
+++ b/access-control.md
@@ -1,46 +1,72 @@
-Operations Overview (subdomain ops are exempted)
-
-High Level Operations
-
-zNSRegistry
-
-zNSDomainToken
-
-zNSAddressResolver (zNS Default Resolver?)
-
-zNS TLD Registrar/SubdomainRegistrar (ZNSEthRegistrar)
-* mint new domains
-* Update name storage of zNSRegistry (TLD only?)
-
-zNSPriceOracle
+** Access Control Spike **
 
 Reference Code https://github.com/zer0-os/zNS/pull/18/
 
-Goal: Identify internal operations and assign them sec, no sec
-Identify admin only operations
-Identify operations of specific users
-Identify operations needed to maintain a subset of the system
+Goal: Identify an appropriate level of access control for zNS administration and its consumers for the current suite of functionality. This does not include subdomain minting.  
 
-Operations
+This analysis assumes a few guiding architecture principles of zNS
+1. Maximize User Ownership:  Operations on entities within zNS are owned (and also the responsibility of) the user. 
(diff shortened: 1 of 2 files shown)


access-control.md
```diff
@@ -0,0 +1,98 @@
...
+    * setDomainRecord - sec - Admin (create)/operator(modify/create)
+        * Admin can initialize a domain with defaults 
+        * Operators are set on creation, with a sole operator defaulting to owner, but with the option to define multiple operators
+        * modification of TLDs can be done via individual functions or with different require checks within this function
+    * setOwner - sec - operator only
+        * setOwner becomes -> clearOperatorsAndSend, requires multisig majority 
+    * setResolver - sec - owner only, but should default to address resolver, imo address resolver should be required so that ownership is tied to Ethereum
+2. zNSDomainToken -
+    * TODO Will the ZNSDomainToken be the addressOf (resolver system) for all inital zNS mvps?
+    * register - sec, zero admin (where admin is ZNSRegistry)
```
Whytecrowe, line 41: commenting on this line, but it applies to many:

This is possibly because our main Registrar PR was not merged at the time of your review or the similar naming confusion happened, but the situation is a bit different - ZNSEthRegistrar is, actually, the contract that performs all of the operations, so it should be this contracts that gets it's own role. I was thinking of just adding a role called `REGISTRAR_ROLE`, that will be assigned to the address of the Registrar contract.

All of the functions that are part of the "atomic operation" like registration, for example, have to be performed together only through the flow triggered by the Registrar. which means, we will have to add modifiers that can only let REGISTRAR_ROLE through to the function. for example: ZNSDomainToken::revoke or ZNSDomainToken::register and similar accessed functions in the Registry such as setDomainResolver/Owner etc. 

But here, the Registry has to be reworked to allow owner or operator to only **modify** a domain, but not create a new one, since all the new domains have to be created through the flow that ZNSEthRegistrar provides, so they should not be accessible directly. e.g. if domain already exists, you can re-set the record, but if it does not, then only Registrar can continue from this point.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+Key
+* nosec = no access control security needed. 
+* sec = access control recommended, will be followed by a list of roles
+
+1. zNSRegistry - Creates new domain records and sets owner(s) and resolver. Controls modification of those records. Allows for querying of zNS domains and their operators by owner.
+    * exists - nosec
+    * isAllowedOperator - nosec,
+    * getDomainRecord -nosec
+    * setDomainRecord - sec - Admin (create)/operator(modify/create)
```
Whytecrowe, line 32: Only Registrar contract access. no one should be able to create a new domain on ZNSRegistry directly, circumventing the staking, token minting and other ops. instead, after all of those operations have been successfully triggered by the RegistRAR, it will trigger the creation of a new domain on RegistRY.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+        * Admin can initialize a domain with defaults 
+        * Operators are set on creation, with a sole operator defaulting to owner, but with the option to define multiple operators
```
Whytecrowe, line 34: I think, operators MAY be set by the owner at a different time if he needs any.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+        * modification of TLDs can be done via individual functions or with different require checks within this function
```
Whytecrowe, line 35: yes. this can be done by every existing domain owner or operator. 
...still not sure what access operators might have that's different from the owner. is it only to add more addresses?...

access-control.md
```diff
@@ -0,0 +1,98 @@
+**Access Control Investigation**
+
+Reference Code https://github.com/zer0-os/zNS/pull/18/
+
+**Goal**
+Identify optimal access control strategies for zNS administratators and consumers for the current suite of functionality. (Subdomains excluded)
+
```
Whytecrowe, line 7: Great start to this and good analysis by function, although we are still missing Treasury.
Below, I will provide my view of things based not on individual operations, but on the whole flows from start to finish.
Hopefully, both of these analyses will yield the result we need.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+1. zNSRegistry - Creates new domain records and sets owner(s) and resolver. Controls modification of those records. Allows for querying of zNS domains and their operators by owner.
+    * exists - nosec
+    * isAllowedOperator - nosec,
+    * getDomainRecord -nosec
+    * setDomainRecord - sec - Admin (create)/operator(modify/create)
+        * Admin can initialize a domain with defaults 
+        * Operators are set on creation, with a sole operator defaulting to owner, but with the option to define multiple operators
+        * modification of TLDs can be done via individual functions or with different require checks within this function
+    * setOwner - sec - operator only
+        * setOwner becomes -> clearOperatorsAndSend, requires multisig majority
```
Whytecrowe, line 37: 1. should it be operator facing? or only current owner facing? My vote is for the latter. only owner should be able to assign a new owner. thoughts?

2. we don't need to clear operators here. they will not be able to access if the modifier/function is written properly. in the current implementation of `'onlyOwnerOrOperator` it already works. operator is the 2nd level key in a mapping, meaning, under the new owner this operator will resolve to `false`.

3. Multisig? What for? Would it be an overkill?

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+    * exists - nosec
+    * isAllowedOperator - nosec,
+    * getDomainRecord -nosec
+    * setDomainRecord - sec - Admin (create)/operator(modify/create)
+        * Admin can initialize a domain with defaults 
+        * Operators are set on creation, with a sole operator defaulting to owner, but with the option to define multiple operators
+        * modification of TLDs can be done via individual functions or with different require checks within this function
+    * setOwner - sec - operator only
+        * setOwner becomes -> clearOperatorsAndSend, requires multisig majority 
+    * setResolver - sec - owner only, but should default to address resolver, imo address resolver should be required so that ownership is tied to Ethereum
```
Whytecrowe, line 38: 1. why not operator? seems like an op that an operator can be responsible for, in case he manages the domain content. this operator may even be a contract for a whale domainer that has 50 domains in ZNS :)

2. should not default to address resolver, cause we are looking to expand. and we can deploy new resolvers at any time, if we tied it to one now, in the future along with deploying a new resolver we would have to also upgrade the registry, which for us, at any point in time, should be a thing we avoid.

3. Not sure why we need to require the address resolver here? What if I make my domain for a has or a string or smth and I don't care about the addresses at all?

4. We eventually want ZNS to be chain agnostic. How possible this is, we will find out later :) but, in theory, we don't have to tie ownership to Ethereum. people just have to understand that our contracts are on ethereum.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+2. zNSDomainToken -
+    * TODO Will the ZNSDomainToken be the addressOf (resolver system) for all inital zNS mvps?
```
Whytecrowe, line 40: Don't understand the question here.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+    * register - sec, zero admin (where admin is ZNSRegistry)
+    * revoke - sec, zero admin (where admin is ZNSRegistry), provide method in registry to initiate revoke process
```
Whytecrowe, line 42: initiator is [ZNSEthRegistrar::revokeDomain()](https://github.com/zer0-os/zNS/blob/development/contracts/distribution/ZNSEthRegistrar.sol#L182)

The Registrar function is ONLY available to the current domain owner.

This function in the DomainToken should only be callable by the ZNSEthRegistrar, since revoking is an atomic operation that starts at the Registrar and has to go all the way through. We shouldn't allow just burning the token without revoking the domain everywhere.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+4. zNSPriceOracle
+    * getPrice - nosec
+    * setBasePrice - sec operator and owner 
+    * setPriceMultiplier - sec - operator and owner
+    * setBaseLength - sec - operator and owner
+    * setBaseLengths (set default price length for domains and subdomains) - sec operator and owner
+        * RFC should default be able to change on subdomains, following principle of democratic user control, up to them
+        * perhaps should allow for a configuration to be set on subdomain systems
+    * setZNSRegistrar ()
+    * Summary: Zero admin should only have control over initalization/upgradeability based on our principles, everything else can be owner
```
Whytecrowe, line 56: to be clear, `owner` and `operator` relate only to the domains. all of these functions are setters of the system storage, so they should only be accessible by the Zero Admin, because we will set prices.

In the future, the plan is to offer an additional way for a domain owner to set their own prices for subdomains, therefore creating an economy on ZNS.

since we are not doing subdomains now, we can avoid all this and limit these functions to the ADMIN_ROLE.

Only PriceOracle getters should be publicly callable. everything else is set by Zero.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+**Conclusions and Recommendations**:
+
+
+zNSPriceOracle - All price setting operation access can be limited to owners & operators. If there is a need to maintain registrar mappings in the price oracle then it may be appropriate for ZAC to assign those mappings. Monetary operations to guarantee revenue of Zero system (e.g. royalties) should be hardcoded and immutable (or at a minimum separate operations controlled by Zero Admins)
```
Whytecrowe, line 61: interesting point on immutable royalties... sounds good, but I'm not sure our management would want that. they would probably want them to be configurable through admin wallet, so - variables.

Regarding the owner/operator see a comment above.

Not sure what "registrar mappings in the price oracle" means...

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+zNSAddressResolver - Resolution of zNS name to a system (setting the contract that acts as the resolver) is data that should be the sole propriety of the owner, or with the understanding that operators are given full trust of a zNS domain. This is a key relationship of zNS, and answers: in what way does this domain name have meaning within ethereum? E.g. is it tied to another system of SCs or is it just a name.
```
Whytecrowe, line 64: agreed. we should not have any access to change user domain resolutions!

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+**Conclusions and Recommendations**:
+
+
+zNSPriceOracle - All price setting operation access can be limited to owners & operators. If there is a need to maintain registrar mappings in the price oracle then it may be appropriate for ZAC to assign those mappings. Monetary operations to guarantee revenue of Zero system (e.g. royalties) should be hardcoded and immutable (or at a minimum separate operations controlled by Zero Admins)
+
+
+zNSAddressResolver - Resolution of zNS name to a system (setting the contract that acts as the resolver) is data that should be the sole propriety of the owner, or with the understanding that operators are given full trust of a zNS domain. This is a key relationship of zNS, and answers: in what way does this domain name have meaning within ethereum? E.g. is it tied to another system of SCs or is it just a name. 
+
+
+zNSRegistry - Similar to zNS Address resolver,modification of records and resolvers require such a high degree of trust that a decision needs to be made on what the purview of operators should be. (See aside on operators and owners for recommendations)
```
Whytecrowe, line 67: yes. we need to decide what operators can and can't do. since they are "operating", maybe they just won't have access to assign a new owner, but can do the rest. maybe they can only create subdomains in the future.
what do you guys think?

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+zNSDomainToken - Because this contract appears to exist as a way to separate minting/revoke concerns from the registry itself, these contracts can be be maintained by the Zero Admin Role. The zNSRegistry contract should be the primary admin in this scenario
```
Whytecrowe, line 70: some yes, some no.

this contract is a tokenization of every domain. standard ERC-721 functions should be accessible as any other NFT, revoke and register should be limited only to ZNSEthRegistrar and if we need to set any important vars on this (if we add any), those should be allowed to set by the Zero Admin.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+**Recommendations on Operators and Owners:**
+
+
+The analysis of the above contracts suggests that nearly every operation (e.g. manipulating price, defining resolvers) granted to an operator in zNS MVP requires such a high degree of trust to the point that operators might as well be owners from a domain operability perspective.
```
Whytecrowe, line 76: good point. need to think on what their differences are in access and roles they play

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+In tandem with our principle of high user responsibility, I recommend we allow operators to perform any operation with the exception of minting, changing domain ownership, and managing operators. This allows operators full bandwidth to help manage a system, but with owners always have the capability to recover from a breach of trust
```
Whytecrowe, line 79: yes, but this leaves only changing resolver address :)

Maybe we give them creation of subdomains and changing resolvers. Changing ownership and managing operators will be for owners only.

OR

We can let them do everything but changing the owner. Might be useful...

Thoughts?

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+Another possible route, which might simplify AC but add complexity elsewhere is to simply allow for multiple owners, and remove the operator role entirely.
```
Whytecrowe, line 82: good thought, but this will be a nightmare to manage

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+Access Control by Zero should be limited to the cases where the user is minting a domain under default contracts that provide some base level of utility in our ecosystem. E.g. Zero Access Control can set a domains resolver to the default if no custom resolver is provided. One these defaults are overriden by the owner. zNS has no access to revoke.
+
+
+While the zNS MVP matures and is in development I can propose two potential paths for admin level access.
+
+
+Option 1:
+ Zero maintains rights to upgradeability over contracts until such a time that Zero Access control decisions are made via DAO or some other multisig. This may provide some assurances to the userbase about our commitment to a decentralized, democratic system, but in reality we are only obscuring centralization that is still there.
```
Whytecrowe, line 94: this is a better option. and the plan of the management. at least, so far.

access-control.md
```diff
@@ -0,0 +1,98 @@
...
+
+Option 2:
+ Zero does all of the above, but an ADMIN role is provided access to manage state variables of our default contracts and TLD registries. This will make iterating on technical upgrades far more efficient at the cost of appearing less decentralized. 
+
```
Whytecrowe, line 98: The main pillars of ZNS AC:

1. Top level (entry point) functions that need to be called by the user should ONLY be callable by the user. If Zero wants to create a domain, it has to go through the exact same process as any other user (staking, minting a token, etc.).

2. Lower level functions for the same flows should be limited only to a specific calling contract, in our case, it's mostly just ZNSEthRegistrar.

3. Functions for changing domain data:
      a. Accessible only to owner/operator for EXISTING domains
      b. Accessible only to caller contract (Registrar) for NEW domains

4. Crucial functions for system integrity: storage setters, initializers, etc. should only be accessible by Zero Admin and all of the levels of roles planned for system governance.

5. Zero Admins should NOT have only 1 role, but have a good structure where we can't lose the system by compromising a single wallet address. We need at lest 2 levels: GOVERNOR, ADMIN. Governor can assign new admins and has access to that + some of the functions, where ADMINs operate the system, by setting storage values, prices, etc. This is just a simple example. Feel free to offer better solutions if you have any in mind.

6. Think on upgradability and what can be added later vs is required from the get go.

Whytecrowe: validated this ^ approach with management. they have the same idea.

PR #22, opened 2023-05-12
fix: Making ZNSDomainToken Upgradeable

Will-Hyde:
Adding the Upgradeable ERC721 interface to the domain token, so that it can make use of Open Zeppelin's Transparent Proxy


JamesEarle (approved):

RyanCross:
@Whytecrowe @Will-Hyde What do y'all think about making this proxy UUPS (inheriting from https://docs.openzeppelin.com/contracts/4.x/api/proxy#UUPSUpgradeable)

OpenZeppelin's latest recommendation on proxies appears to have shifted over time to this pattern.  Which moves upgrade logic to the implementation contract and not the proxy contract.

Further reading:
https://docs.openzeppelin.com/contracts/4.x/api/proxy#transparent-vs-uups

Whytecrowe:
> @Whytecrowe @Will-Hyde What do y'all think about making this proxy UUPS (inheriting from https://docs.openzeppelin.com/contracts/4.x/api/proxy#UUPSUpgradeable)
> 
> OpenZeppelin's latest recommendation on proxies appears to have shifted over time to this pattern. Which moves upgrade logic to the implementation contract and not the proxy contract.
> 
> Further reading: https://docs.openzeppelin.com/contracts/4.x/api/proxy#transparent-vs-uups

Could you, please, look into it @JamesEarle ? Maybe UUPS would be better here?

contracts/token/ZNSDomainToken.sol
```diff
@@ -1,60 +1,57 @@
...
+     * @notice Set the address of the access controller contract
+     * @param _accessController The new access controller contract
+     */
+    function setAccessController(address _accessController) external override onlyRole(GOVERNOR_ROLE) {
+        _setAccessController(_accessController);
+    }
+
+    /**
+     * @notice To use UUPS proxy we override this function and revert if `msg.sender` isn't authorized
+     * @dev Using solhint's `no-empty-blocks` will error here, but to be a UUPS Proxy we require it this
```
Whytecrowe, line 49: looks like an outdated comment

package.json
```diff
@@ -11,16 +11,16 @@
   },
   "scripts": {
     "compile": "hardhat compile",
-    "lint-sol": "yarn solhint \"contracts/**/*.sol\"",
+    "lint-sol": "yarn solhint ./contracts/**/*.sol",
     "lint-ts": "yarn eslint ./test/** ./src/**",
-    "lint": "yarn lint-sol && yarn lint-ts",
+    "lint": "yarn lint-sol & yarn lint-ts",
```
Whytecrowe, line 16: you want these to run in parallel?

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+        ethers.constants.AddressZero,
+        caller.address,
         tokenId
       );
-      expect(receipt.events?.[0].args?.to).to.eq(
-        caller.address
-      );
-
-      // Verify caller owns tokenId
-      expect(await domainToken.ownerOf(tokenId)).to.equal(caller.address);
```
Whytecrowe, line 42: why is this removed? we need to verify that the token was, in fact, transferred to the correct address. an event does not verify this

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+        caller.address,
+        ethers.constants.AddressZero,
         tokenId
       );
-      expect(receipt.events?.[0].args?.to).to.eq(
-        ethers.constants.AddressZero
-      );
-
-      // Verify token has been burned
-      await expect(
```
Whytecrowe, line 70: same here

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
 
   describe("Require Statement Validation", () => {
+    it("Only the registrar can call to register a token", async () => {
+      const tokenId = ethers.BigNumber.from("1");
+      const registerTx = zns.domainToken
+        .connect(caller)
+        .register(caller.address, tokenId);
+
+      await expect(registerTx).to.be.revertedWith(
+        `AccessControl: account ${caller.address.toLowerCase()} is missing role ${REGISTRAR_ROLE}`
```
Whytecrowe, line 83: there's a helper for this:
https://github.com/zer0-os/zNS/blob/development/test/helpers/errors.ts#L1

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
     });
 
     it("Verify token symbol", async () => {
-      const symbol = await domainToken.symbol();
+      const symbol = await zns.domainToken.symbol();
       expect(symbol).to.equal(TokenSymbol);
     });
   });
+  describe("AccessController", () =>{
+    it("Allows setting of a new access controller if the caller is a governor", async () => {
```
Whytecrowe, line 124: admin, not governor

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+      await expect(tx).to.not.be.reverted;
+
+      const accessControllerAfter = await zns.domainToken.getAccessController();
+      expect(accessControllerBefore).to.not.eq(accessControllerAfter);
+    });
+
+    it("Fails when the caller is not a governor", async () => {
+      const tx = zns.domainToken.connect(caller).setAccessController(mockAccessController.address);
+      await expect(tx).to.be.revertedWith(
+        `AccessControl: account ${caller.address.toLowerCase()} is missing role ${GOVERNOR_ROLE}`
```
Whytecrowe, line 138: use helper

test/ZNSDomainToken.test.ts
Whytecrowe: also, the newly merged AC PR has all these tests. please, verify and remove redundant ones

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+      );
+    });
+  });
+  describe("UUPS", () => {
```
Whytecrowe, line 142: space before this

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+
+      const preUpgradeVars = [
+        zns.domainToken.name(),
+        zns.domainToken.symbol(),
+      ];
+
+      const [nameBefore, symbolBefore] = await Promise.all(preUpgradeVars);
+
+      // Confirm the deployer is a governor, as set in `deployZNS` helper
+      await expect(zns.accessController.checkRole(GOVERNOR_ROLE, deployer.address)).to.not.be.reverted;
```
Whytecrowe, line 160: use `hasRole()` or `isGovernor()` instead here. they return a boolean.

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+
+      // Confirm the deployer is a governor, as set in `deployZNS` helper
+      await expect(zns.accessController.checkRole(GOVERNOR_ROLE, caller.address)).to.be.revertedWith(
+        `AccessControl: account ${caller.address.toLowerCase()} is missing role ${GOVERNOR_ROLE}`
+      );
+
+      const upgradeTx = zns.domainToken.connect(caller).upgradeTo(newDomainToken.address);
+
+      await expect(upgradeTx).to.be.revertedWith(
+        `AccessControl: account ${caller.address.toLowerCase()} is missing role ${GOVERNOR_ROLE}`
```
Whytecrowe, line 195: helper

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+
+      const postUpgradeVars = [
+        zns.domainToken.name(),
+        zns.domainToken.symbol(),
+      ];
+
+      const [nameAfter, symbolAfter] = await Promise.all(postUpgradeVars);
+
+      expect(nameBefore).to.eq(nameAfter);
+      expect(symbolBefore).to.eq(symbolAfter);
```
Whytecrowe, line 174: what about the state vars that actually matter? :) like token ownership and stuff...
what about the new functionality? is it added with the upgrade? are new slots added with the upgrade? do they interfere with old slots?

Whytecrowe: a helper can be potentially created here that can be reused for tests of other proxies.

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+      await expect(tx).to.be.revertedWith(
+        `AccessControl: account ${caller.address.toLowerCase()} is missing role ${GOVERNOR_ROLE}`
+      );
+    });
+  });
+  describe("UUPS", () => {
+    it("Verifies an authorized user can upgrade the contract", async () => {
+      // UUPS specifies that a call to upgrade must be made through an address that is upgradecall
+      // So use a deployed proxy contract
+      const factory = new ZNSDomainToken__factory(deployer);
```
Whytecrowe, line 146: need to upgrade to a contract with slightly different logic (new function(-s), new storage slots) to actually verify. upgrading to the identical logic doesn't really check much

test/ZNSDomainToken.test.ts
```diff
@@ -14,101 +15,185 @@ describe("ZNSDomainToken:", () => {
...
+
+      expect(nameBefore).to.eq(nameAfter);
+      expect(symbolBefore).to.eq(symbolAfter);
+
+    });
+
+    it("Fails to upgrade if the caller is not authorized", async () => {
+      // UUPS specifies that a call to upgrade must be made through an address that is upgradecall
+      // So use a deployed proxy contract
+      const factory = new ZNSDomainToken__factory(deployer);
```
Whytecrowe, line 181: for the AC failed call, this is fine

test/helpers/deployZNS.ts
```diff
@@ -69,10 +69,23 @@ export const deployPriceOracle = async ({
...
+  deployer : SignerWithAddress,
+  accessController : string
 ) : Promise<ZNSDomainToken> => {
   const domainTokenFactory = new ZNSDomainToken__factory(deployer);
-  return domainTokenFactory.deploy("ZNSDomainToken", "ZDT");
+  const contract = await upgrades.deployProxy(
+    domainTokenFactory,
+    [
+      accessController,
+      "ZNSDomainToken",
```
Whytecrowe, line 80: these should be constants, especially if we check them in multiple tests and use them in helpers

test/helpers/deployZNS.ts
```diff
@@ -139,12 +152,12 @@ export const deployZNS = async ({
     adminAddresses: [deployer.address, ...adminAddresses],
   });
 
-  // TODO AC: fix this!
+  // TODO AC: Make sure contracts are deployed as proxies and authorize the governor role
```
Whytecrowe, line 155: not sure what this comment means

Whytecrowe (changes requested):
please address the comments + resolver conflicts

package.json
JamesEarle: I was getting different output if I ran in parallel vs consecutively, so I left it.

test/ZNSDomainToken.test.ts
JamesEarle: Meant to put this back, it's added

test/ZNSDomainToken.test.ts
JamesEarle: changed all of them to this, thanks

test/helpers/deployZNS.ts
JamesEarle: a lot clearer than comments like "fix this!" to me, but this was because some contracts were being deployed through their factory still not through the `upgrades.deployProxy()` command like others, but this has been changed

PR #24, opened 2023-05-17
fix: Rework registry for only top level domains

JamesEarle:
Have not yet updated tests.

Some file changes are just formatting to 4 space indentation


.solhint.json
```diff
@@ -1,16 +1,28 @@
...
+    "func-visibility": [
+      "warn",
+      {
+        "ignoreConstructors": true
+      }
+    ],
     "constructor-syntax": "warn",
-    "named-parameters-mapping": "error"
+    "named-parameters-mapping": "error",
+    "indent": [
```
Whytecrowe: in the newest solhint this rule has been removed

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -1,252 +1,166 @@
...
-
-      // We remove subdomain approval immediately after it is used
-      setSubdomainApproval(parentDomainHash, msg.sender, false);
+import {IZNSEthRegistrar} from "./IZNSEthRegistrar.sol";
+import {IZNSRegistry} from "../registry/IZNSRegistry.sol";
+import {IZNSTreasury} from "./IZNSTreasury.sol";
+import {IZNSDomainToken} from "../token/IZNSDomainToken.sol";
+import {IZNSAddressResolver} from "../resolver/IZNSAddressResolver.sol";
+import {IZNSPriceOracle} from "./IZNSPriceOracle.sol";
+import {StringUtils} from "../utils/StringUtils.sol";
```
Whytecrowe: where did all the spaces go?
I would really recommend not using prettier, since it will keep messing our things up

JamesEarle: I use the Solidity extension in VS code for all the niceties in development, and by default it uses prettier. The other option for it is forge. I can remove the .prettierrc file but we should discuss how to properly have the same formatting setup for everything for sure.

contracts/distribution/IZNSEthRegistrar.sol
```diff
@@ -21,19 +21,22 @@ interface IZNSEthRegistrar {
...
         address resolver,
         address domainContent
     ) external returns (bytes32);
 
-    function setSubdomainApproval(
-        bytes32 parentHash,
-        address user,
-        bool status
-    ) external;
+    // function setSubdomainApproval(
```
Whytecrowe: to remove

contracts/distribution/ZNSEthRegistrar.sol
Whytecrowe: we should, yeah.
I am probably the minority here, but my hands and a vision of the code in my head is usually enough to just split lines and format as I'm writing. every time I use prettier, it never actually formats the way I want + the lack of support for destructuring really annoys me... :))

Whytecrowe: A good rule I'm using: I don't like to move mouse around to highlight the code. I usually click on the line number to highlight, so every time I highlight by line it should be one logical step always. it matches with good code formatting most of the time.

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -1,28 +1,32 @@
...
     IZNSDomainToken public znsDomainToken;
     IZNSAddressResolver public znsAddressResolver;
     IZNSPriceOracle public znsPriceOracle;
 
+    // To account for unicode strings, we need a custom length library
+    using StringUtils for string;
 
-    mapping(bytes32 parentDomainHash => mapping(address user => bool status))
-    public subdomainApprovals;
+    mapping(bytes32 parentdomainNameHash => mapping(address user => bool status))
```
Whytecrowe: why break camelCase here?

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -45,208 +49,118 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
         address resolver,
-        address domainAddress
+        address resolverContent
     ) external returns (bytes32) {
-        require(bytes(name).length != 0, "ZNSEthRegistrar: Domain Name not provided");
-
-        // To not repeat external calls, we load into memory
-        bytes32 rootHash = znsRegistry.ROOT_HASH();
+        require(
+            name.strlen() != 0,
```
Whytecrowe: is it worth using the lib here? it's not cheap to calc length this way. is there a case where the bytes count is not going to be correct for 0 comparison here?

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -45,208 +49,118 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
+            "ZNSEthRegistrar: Domain Name not provided"
+        );
 
         // Create hash for given domain name
-        // TODO: remove hashing with the rootHash and remove root hash completel!
-        bytes32 domainHash = hashWithParent(rootHash, name);
+        bytes32 domainNameHash = keccak256(bytes(name));
```
Whytecrowe: is "Name" really necessary here? :) we have this name different in different contracts.

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -45,208 +49,118 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
-        // value vs. just address(0)? What are pros and cons?
-        emit DomainRegistered(
-            rootHash,
-            domainHash,
-            tokenId,
-            name,
-            msg.sender,
-            resolver
-        );
+        // emit DomainRegistered(domainNameHash, tokenId, name, msg.sender, resolver);
```
Whytecrowe: ?

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -45,208 +49,118 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
-            );
-
-            registerFor = msg.sender;
-
-            // We remove subdomain approval immediately after it is used
-            setSubdomainApproval(parentDomainHash, msg.sender, false);
-        }
+    // TODO Functions for updating the domain already exist in ZNSRegistry
+    // We can have the dApp call those functions directly?
+    // Should we add an `updateDomain` function here instead?
```
Whytecrowe: if we need to change owner or resolver for an existing domain, we can just call Registry.

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -45,208 +49,118 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
+    ) external onlyOwner(domainNameHash) {
+        uint256 tokenId = uint256(domainNameHash);
         znsDomainToken.revoke(tokenId);
-        znsTreasury.unstakeForDomain(domainHash, msg.sender);
-        znsRegistry.deleteRecord(domainHash);
+        znsTreasury.unstakeForDomain(domainNameHash, msg.sender);
+        znsRegistry.deleteRecord(domainNameHash);
 
-        emit DomainRevoked(domainHash, msg.sender);
+        // emit DomainRevoked(domainNameHash, msg.sender);
```
Whytecrowe: don't forget to uncomment this

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -45,208 +49,118 @@ contract ZNSEthRegistrar is IZNSEthRegistrar {
...
-                domainHash,
-                owner,
-                resolver
-            );
+            znsRegistry.setDomainRecord(domainNameHash, owner, resolver);
+
             // TODO error: we are using a different Resolver here than the one passed!
             //  and the one in the call above. This is an error that can cause undiscoverable domains
             //  If we only have one AddressResolver, we should not take it as an argument
             //  to the register function at all and assume the one in this storage is being used
```
Whytecrowe: this was not addressed here. we are still using 2 different resolvers in these 2 calls without any validation.
we have 2 options here:
1. not having `resolver` as param to the `registerDomain()` and just using the one from local storage
2. keep the `resolver` param, but check if it's the the same one we have in storage (this is, imo, not necessary, since we should have only one AddressResolver anyway.)

contracts/registry/IZNSRegistry.sol
```diff
@@ -87,98 +85,78 @@ interface IZNSRegistry {
...
+     *
+     * @param domainNameHash The hash of the domain
+     * @param owner The owner or an allowed operator of that domain
+     * @param resolver The resolver for the domain
      */
-    function setSubdomainOwner(
-        bytes32 parentDomainHash,
-        bytes32 domainHash,
-        address owner
+    function setDomainRecord(
```
Whytecrowe: maybe "change" or "modify" instead of "set" here? "set" usually means set the value in any case. here we need it to only work on the existing record and it be clear in the name.

contracts/registry/ZNSRegistry.sol
```diff
@@ -1,60 +1,81 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.18;
 
-import { ERC1967UpgradeUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/ERC1967/ERC1967UpgradeUpgradeable.sol";
-import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
-import { IZNSRegistry } from "./IZNSRegistry.sol";
+import {ERC1967UpgradeUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/ERC1967/ERC1967UpgradeUpgradeable.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import {IZNSRegistry} from "./IZNSRegistry.sol";
```
Whytecrowe: spaces disappeared again

contracts/registry/ZNSRegistry.sol
```diff
@@ -65,6 +86,22 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
         return candidate == owner || operators[owner][candidate];
     }
 
+    /**
+     * @notice Change the address of the ZNSRegistrar contract we use
+     *
+     * @param znsRegistrar_ The new ZNSRegistrar
+     */
+    // TODO When we have access control, only be callable by admin!!
```
Whytecrowe, line 92: with access control, we won't need to store registrar address on here at all and all this logic will go away. we will just use the AC contract to check the role of the caller being only REGISTRAR_ROLE

contracts/registry/ZNSRegistry.sol
```diff
@@ -78,128 +115,125 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
...
-    function getDomainOwner(
-        bytes32 domainHash
-    ) external view returns (address) {
-        return records[domainHash].owner;
+    function setDomainRecord(
+        bytes32 domainNameHash,
+        address owner,
+        address resolver
+    ) external onlyOwnerOrOperator(domainNameHash) {
+        require(_exists(domainNameHash), "ZNSRegistry: Domain does not exist");
```
Whytecrowe: are we ever going to hit this line if a domain doesn't exist? wouldn't it always fail in the modifier?

contracts/registry/ZNSRegistry.sol
```diff
@@ -78,128 +115,125 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
...
-    //   _setDomainOwner(domainHash, owner);
-
-    //   // TODO probably don't need any "domain" functions
-    //   // emit DomainOwnerSet(owner, domainNameHash);
-    // }
+    function setDomainOwner(
+        bytes32 domainNameHash,
+        address owner
+    ) external onlyOwnerOrOperator(domainNameHash) {
+        require(_exists(domainNameHash), "ZNSRegistry: Domain does not exist");
```
Whytecrowe: same here

contracts/registry/ZNSRegistry.sol
```diff
@@ -78,128 +115,125 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
...
+     *
+     * @param domainNameHash the hash of a domain's name
      * @param resolver The new default resolver
      */
     function setDomainResolver(
         bytes32 domainNameHash,
         address resolver
-    ) public onlyOwnerOrOperator(domainNameHash) {
+    ) external onlyOwnerOrOperator(domainNameHash) {
+        require(_exists(domainNameHash), "ZNSRegistry: Domain does not exist");
```
Whytecrowe: same here. double-check that this line is needed

contracts/registry/ZNSRegistry.sol
```diff
@@ -78,128 +115,125 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
...
     }
 
+    /**
+     * @notice Delete a domain's record
+     *
+     * @param domainNameHash The hash of the domain name
+     */
+    function deleteRecord(bytes32 domainNameHash) external onlyRegistrar {
+        // TODO Test if after revocation an operator can do anything to verify
+        // we don't need to clear them.
```
Whytecrowe: you talking about this? ->

https://github.com/zer0-os/zNS/blob/proto/access-control/test/ZNSEthRegistrar.test.ts#L711

contracts/registry/ZNSRegistry.sol
```diff
@@ -221,14 +255,21 @@ contract ZNSRegistry is IZNSRegistry, ERC1967UpgradeUpgradeable {
...
      * @param resolver The resolver to set
      */
     function _setDomainResolver(
         bytes32 domainNameHash,
         address resolver
     ) internal {
-        require(resolver != address(0), "ZNSRegistry: Resolver can NOT be zero address");
+        // TODO Allow setting resolver to 0?
+        // If owner is 0, it is a burned domain
+        // If resolver is zero, nothing happens
```
Whytecrowe: what would be the case where we would want to set it to 0?

contracts/resolver/ZNSAddressResolver.sol
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.18;
 
-import { ERC165 } from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
-import { IZNSAddressResolver } from "./IZNSAddressResolver.sol";
-import { IZNSRegistry } from "../registry/IZNSRegistry.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {IZNSAddressResolver} from "./IZNSAddressResolver.sol";
+import {IZNSRegistry} from "../registry/IZNSRegistry.sol";
```
Whytecrowe: spaces again

contracts/registry/ZNSRegistry.sol
```diff
@@ -1,60 +1,81 @@
...
         _;
     }
 
     /**
      * Initialize the ZNSRegistry contract, setting the owner of the `0x0` domain
      * to be the account that deploys this contract
      */
-    function initialize(address owner) public initializer {
-        require(owner != address(0), "ZNSRegistry: Owner can not be 0x0 address");
-        records[ROOT_HASH].owner = owner;
```
Whytecrowe, line 41: please, add some tests to make sure removal of this is not going to be a problem. not sure yet how to test it properly.
the problem might arise when we add subdomains..?...

contracts/distribution/ZNSEthRegistrar.sol
JamesEarle: unintentional, fixed it!

contracts/distribution/ZNSEthRegistrar.sol
JamesEarle: Good point, resetting to `bytes(name).length` instead

contracts/distribution/ZNSEthRegistrar.sol
JamesEarle: Felt like it's more clear, but I'm impartial

contracts/distribution/ZNSEthRegistrar.sol
JamesEarle: Temporary modifications while I was debugging something, I've uncommented this and other events in this file

contracts/registry/ZNSRegistry.sol
JamesEarle: yes it will, so I removed the exists check because we can rely on this

contracts/registry/ZNSRegistry.sol
JamesEarle: I'm also unsure how to test, what would the potential "problems" caused by this removal be? For now, we won't have subdomains and so we don't need an owner of a parent domain.

Whytecrowe (approved):

PR #26, opened 2023-05-22
fix: make ZNSPriceOracle a UUPS proxy

JamesEarle:


contracts/distribution/ZNSPriceOracle.sol
```diff
@@ -202,4 +203,14 @@ contract ZNSPriceOracle is AccessControlled, IZNSPriceOracle, Initializable {
...
+
+    /**
+     * @notice To use UUPS proxy we override this function and revert if `msg.sender` isn't authorized
+     * @dev Using solhint's `no-empty-blocks` will error here, but to be a UUPS Proxy we require it this
+     * and so we simply disable solhint for this function
+     * 
+     * @param newImplementation The new implementation contract to upgrade to.
+     */
+     // solhint-disable-next-line no-empty-blocks
+    function _authorizeUpgrade(address newImplementation) internal override onlyRole(GOVERNOR_ROLE) {}
```
Whytecrowe, line 215: we should do this in the same way in all contracts. in the Token, the access check is in the body of the function.

test/ZNSPriceOracle.test.ts
```diff
@@ -579,4 +580,64 @@ describe("ZNSPriceOracle", () => {
...
+      const proxyPriceOracle = await hre.upgrades.deployProxy(factory, [
+        zns.accessController.address,
+        priceConfigDefault,
+        registrationFeePercDefault,
+      ]);
+
+      await proxyPriceOracle.deployed();
+
+      // PriceOracle to upgrade to
+      const newPriceOracle = await factory.deploy();
```
Whytecrowe, line 597: same here as in the Token PR.
- more important storage slots should be populated and checked after the upgrade
- the contract to upgradeTo should have new functionality and state vars
- the new functionality should be checked along with the old one after the upgrade

You can make mocked updated contracts in the `test` dir that are used just for tests like these.
just add arbitrary function like `sum(a, b)` that writes into new storage slots.

test/ZNSPriceOracle.test.ts
```diff
@@ -579,4 +580,64 @@ describe("ZNSPriceOracle", () => {
...
+        registrationFeePercDefault
+      );
+
+      // Confirm the account is not a governor
+      await expect(zns.accessController.checkRole(GOVERNOR_ROLE, randomAcc.address)).to.be.reverted;
+
+      const tx = proxyPriceOracle.connect(randomAcc).upgradeTo(newPriceOracle.address);
+
+      await expect(tx).to.be.revertedWith(
+        `AccessControl: account ${randomAcc.address.toLowerCase()} is missing role ${GOVERNOR_ROLE}`
```
Whytecrowe, line 639: use helper

PR #29, opened 2023-05-25
feat: make ZNSRegistry a UUPS proxy

JamesEarle:


contracts/access/AccessControlled.sol
```diff
@@ -2,20 +2,34 @@
...
 
     IZNSAccessController internal accessController;
 
-
     modifier onlyAdmin() {
         accessController.checkAdmin(msg.sender);
         _;
     }
 
+    modifier onlyGovernor() {
```
Whytecrowe: probably don't need these. It will overload the size of every contract. we have these on access controller already. should be enough

contracts/registry/ZNSRegistry.sol
```diff
@@ -45,17 +42,12 @@ contract ZNSRegistry is AccessControlled, ERC1967UpgradeUpgradeable, IZNSRegistr
         _;
     }
 
-    modifier onlyRegistrar() {
```
Whytecrowe, line 48: leave this

contracts/registry/ZNSRegistry.sol
```diff
@@ -201,11 +193,11 @@ contract ZNSRegistry is AccessControlled, ERC1967UpgradeUpgradeable, IZNSRegistr
 
     function setAccessController(
         address accessController
-    ) external override(AccessControlled, IZNSRegistry) onlyAdmin {
```
Whytecrowe, line 204: why did you remove this?

contracts/registry/IZNSRegistry.sol
```diff
@@ -152,12 +152,4 @@ interface IZNSRegistry {
      * @param domainHash The hash of the domain name
      */
     function deleteRecord(bytes32 domainHash) external;
-
-    /**
-     * @notice Set the access controller contract
-     * @param accessController The new access controller
-     */
-    function setAccessController(address accessController) external;
```
Whytecrowe, line 160: why did you remove these? they will be inaccessible, no?

Whytecrowe: dangerous thing you are doing. please, let me know why

test/ZNSRegistry.test.ts
```diff
@@ -275,34 +260,34 @@ describe("ZNSRegistry Tests", () => {
...
-      const tx = registry.updateDomainRecord(domainHash, mockResolver.address, ethers.constants.AddressZero);
+      await zns.registry.connect(mockRegistrar).createDomainRecord(domainHash, deployer.address, mockResolver.address);
+      const tx = zns.registry.updateDomainRecord(domainHash, mockResolver.address, ethers.constants.AddressZero);
 
       await expect(tx).to.be.fulfilled;
     });
 
-    it("Cannot update a domain owner if owner is zero address", async () => {
-      const tx = registry
+    it("Cannot set a domain owner if owner is zero address", async () => {
```
Whytecrowe: why are you changing the wording back? I specifically changed to "update" cause that's what we're doing. the function on Registry is called `upadateDomainOwner()`. why making it more confusing?
Update means changing existing, set means setting a value in any case. "set" is very confusing here

test/ZNSRegistry.test.ts
```diff
@@ -312,21 +297,21 @@ describe("ZNSRegistry Tests", () => {
       await expect(tx).to.be.revertedWith(OWNER_NOT_ZERO_REG_ERR);
     });
 
-    it("Can update a domain resolver if resolver is zero address", async () => {
-      await registry
+    it("Can set a domain resolver if resolver is zero address", async () => {
```
Whytecrowe: it's **update**DomainResolver()

test/ZNSRegistry.test.ts
```diff
@@ -312,21 +297,21 @@ describe("ZNSRegistry Tests", () => {
...
 
-      const zeroResolver = await registry.getDomainResolver(wilderDomainHash);
+      const zeroResolver = await zns.registry.getDomainResolver(wilderDomainHash);
 
       expect(zeroResolver).to.be.eq(ethers.constants.AddressZero);
     });
 
-    it("Fails to set a record when caller is not owner", async () => {
-      const tx = registry
+    it("Fails to set a record when caller is not owner or operator", async () => {
```
Whytecrowe: "update" should be here as well

test/ZNSRegistry.test.ts
```diff
@@ -459,20 +449,151 @@ describe("ZNSRegistry Tests", () => {
...
+        accessControllerAfter,
+      ] = await registryPromises(zns.registry, domainHash, [deployer, operator]);
+
+      expect(ownerStatusBefore).to.eq(ownerStatusAfter);
+      expect(operatorStatusBefore).to.eq(operatorStatusAfter);
+      expect(existanceBefore).to.eq(existanceAfter);
+      expect(accessControllerBefore).to.eq(accessControllerAfter);
+    });
+
+    it("Fails when an unauthorized account tries to call to upgrade", async () => {
```
Whytecrowe: this tests's title contradicts the test itself

test/helpers/checkUpgrades.ts
```diff
@@ -0,0 +1,64 @@
...
+  ];
+
+  return Promise.all(calls);
+};
+
+export const upgradedRegistryActions = async (
+  upgradedRegistry : ZNSRegistryMock,
+  newNumber : BigNumber,
+  randomUser : SignerWithAddress
+) => {
```
Whytecrowe: maybe we can change the helpers to a different style:
we can make these more abstract to work with any contract, and this will check values from arg1 to the values from arg2, both of which can be contract specific callbacks that we pass as an array for each arg.
we specify the calls as args in the test itself, this way it will be easily readable of what exactly we are checking in the test, while the helper will make these calls and do comparisons internally. we don't need to see comparisons in the tests as much as what we are checking exactly, which we can tell by the callbacks we pass to the helper in the test body.
what do you think?

contracts/registry/IZNSRegistry.sol
JamesEarle: @Whytecrowe what makes this dangerous? We are inheriting it twice, once from the `AccessControlled` contract and then again in the interface for this contract. I was removing the second reference as it seems unnecessary to do this twice, Why do we need this? It's already `virtual` in the `AccessControlled` contract we inherit from so there isn't a possibility of not overriding it as it will error if we don't

test/ZNSRegistry.test.ts
JamesEarle: moved back to update

test/ZNSRegistry.test.ts
JamesEarle: I had mistakenly put the negative test case in the positive test and vice versa, I've already reversed this in the latest however

contracts/registry/ZNSRegistry.sol
JamesEarle: See below comment, what is the need for inheriting this twice if it's already enforced by `AccessControlled` ?

contracts/access/AccessControlled.sol
```diff
@@ -2,15 +2,14 @@
 pragma solidity ^0.8.18;
 
 import { IZNSAccessController } from "./IZNSAccessController.sol";
-
+import { ZNSRoles } from "./ZNSRoles.sol";
```
Whytecrowe: why is this imported? we don't need this anymore

contracts/distribution/IZNSEthRegistrar.sol
```diff
@@ -28,22 +28,30 @@ interface IZNSEthRegistrar {
...
-    function setAddressResolver(address znsAddressResolver_) external;
+    function setAddressResolver(address addressResolver_) external;
 
     function setAccessController(address accessController_) external;
 
     function getAccessController() external view returns (address);
+
+    function initialize(
+        address accessController_,
+        address znsRegistry_,
```
Whytecrowe, line 52: should probably remove "zns" from here as well then

contracts/distribution/IZNSEthRegistrar.sol
```diff
@@ -28,22 +28,30 @@ interface IZNSEthRegistrar {
...
+        address znsTreasury_,
+        address znsDomainToken_,
+        address znsAddressResolver_
```
Whytecrowe: should probably remove "zns" from here as well then

contracts/distribution/IZNSTreasury.sol
```diff
@@ -52,4 +52,11 @@ interface IZNSTreasury {
     function setAccessController(address accessController) external;
 
     function getAccessController() external view returns (address);
+
+    function initialize(
+        address accessController_,
+        address znsPriceOracle_,
```
Whytecrowe: same here

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -94,14 +93,12 @@ contract ZNSEthRegistrar is AccessControlled, IZNSEthRegistrar {
...
-    function revokeDomain(bytes32 domainHash)
-    external
-    override
     // TODO: figure out how to guard this so people can stake tokens
     //  without the risk of staking contract or wallet to call reclaim+revoke
     //  from underneath them
+    function revokeDomain(bytes32 domainHash) external override 
     onlyNameOwner(domainHash)
-    onlyTokenOwner(domainHash)
+    onlyTokenOwner(domainHash)
```
Whytecrowe: why space?

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -111,60 +108,56 @@ contract ZNSEthRegistrar is AccessControlled, IZNSEthRegistrar {
         emit DomainRevoked(domainHash, msg.sender);
     }
 
-    function reclaimDomain(bytes32 domainHash)
-    external
-    override
+    function reclaimDomain(bytes32 domainHash) external override
     onlyTokenOwner(domainHash)
```
Whytecrowe: if we newline, let's newlineeverything. here only modifier is newlined. let's keep to the recommended syntax for Solidity - revert back to previous or put everything in one line.

contracts/distribution/ZNSEthRegistrar.sol
```diff
@@ -111,60 +108,56 @@ contract ZNSEthRegistrar is AccessControlled, IZNSEthRegistrar {
...
-        emit AddressResolverSet(znsAddressResolver_);
+        emit AddressResolverSet(addressResolver_);
     }
 
-    function setAccessController(address accessController_)
-    external
-    override(AccessControlled, IZNSEthRegistrar)
-    onlyAdmin
+    function setAccessController(address accessController_) external override(AccessControlled, IZNSEthRegistrar) 
+        onlyAdmin
```
Whytecrowe: again here. why did you bring the visibility keywords up, but not the modifier? what's wrong with the conventional approach that's already been there? also, why is modifier indented? this is "dirty syntax", let's revert back to how it was.
and also, let's not do this again, please. we're wasting time reformatting the same thing over and over and commenting on it. there's a reason why I formatted it this way, no need to change it

contracts/distribution/ZNSPriceOracle.sol
```diff
@@ -77,6 +77,8 @@ contract ZNSPriceOracle is AccessControlled, Initializable, IZNSPriceOracle {
 
         fee = getRegistrationFee(domainPrice);
         totalPrice = domainPrice + fee;
+
+        return (totalPrice, domainPrice, fee);
```
Whytecrowe: let's do one or the other method. and not half both. either remove this line or specify naming locally. imo, it was already fine

contracts/distribution/ZNSTreasury.sol
```diff
@@ -63,11 +61,11 @@ contract ZNSTreasury is AccessControlled, IZNSTreasury {
...
-      // Transfer stake amount and fee
-      stakingToken.transferFrom(depositor, address(this), stakeAmount);
-      // TODO make sure we show the approval process to the user here to avoid failed transfer
-      // TODO can we make it so it needs a single approval only?!
-      stakingToken.transferFrom(depositor, zeroVault, registrationFee);
+        // Transfer stake amount and fee
+        stakingToken.transferFrom(depositor, address(this), stakeAmount);
+        // TODO make sure we show the approval process to the user here to avoid failed transfer
+        // TODO can we make it so it needs a single approval only?!
+        stakingToken.transferFrom(depositor, zeroVault, registrationFee);
```
Whytecrowe: remove whitespace at the end

contracts/registry/IZNSRegistry.sol
Whytecrowe: either remove everywhere or leave this one as well.

contracts/registry/ZNSRegistry.sol
```diff
@@ -131,7 +128,7 @@ contract ZNSRegistry is AccessControlled, ERC1967UpgradeUpgradeable, IZNSRegistr
         bytes32 domainHash,
         address owner,
         address resolver
-    ) external override onlyRegistrar {
+    ) external override onlyRegistrar() {
```
Whytecrowe: no parentheses needed

contracts/registry/ZNSRegistry.sol
Whytecrowe: see the previous comment

contracts/registry/ZNSRegistry.sol
```diff
@@ -201,11 +198,11 @@ contract ZNSRegistry is AccessControlled, ERC1967UpgradeUpgradeable, IZNSRegistr
 
     function setAccessController(
         address accessController
-    ) external override(AccessControlled, IZNSRegistry) onlyAdmin {
+    ) external override onlyAdmin {
         _setAccessController(accessController);
     }
 
-    function getAccessController() external view override(AccessControlled, IZNSRegistry) returns (address) {
+    function getAccessController() external view override returns (address) {
```
Whytecrowe: same here

contracts/token/IZNSDomainToken.sol
```diff
@@ -1,12 +1,32 @@
...
+    /**
+     * @notice Mints a token with a specified tokenId, using _safeMint, and sends it to the given address
+     * @param to The address that will recieve the newly minted domain token
+     * @param tokenId The TokenId that the caller wishes to mint/register
+     */
     function register(address to, uint256 tokenId) external;
 
+    /**
+     * @notice Burns the token with the specified tokenId
+     * @dev TODO: Add onlyRole(REGISTRAR_ROLE)
```
Whytecrowe: outdated comment

contracts/token/ZNSDomainToken.sol
```diff
@@ -1,27 +1,27 @@
...
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import { IZNSDomainToken } from "./IZNSDomainToken.sol";
 import { AccessControlled } from "../access/AccessControlled.sol";
 
 
 /**
  * @title A contract for tokenizing domains under ZNS
  */
-contract ZNSDomainToken is AccessControlled, ERC721, IZNSDomainToken {
+contract ZNSDomainToken is AccessControlled, UUPSUpgradeable, ERC721Upgradeable, IZNSDomainToken {
```
Whytecrowe, line 12: think on the order of inheritance here. do we want all the uups slots be before all the token slots? what about the gaps?

contracts/token/ZNSDomainToken.sol
```diff
@@ -1,27 +1,27 @@
...
-
-    constructor(
+    function initialize(
+        address accessController,
         string memory tokenName,
-        string memory tokenSymbol,
-        address accessController
-    ) ERC721(tokenName, tokenSymbol) {
+        string memory tokenSymbol
+    ) public override initializer {
```
Whytecrowe: why it's public here, but external in other contracts? let's bring them to the same visibility everywhere

contracts/token/ZNSDomainToken.sol
```diff
@@ -38,11 +38,11 @@ contract ZNSDomainToken is AccessControlled, ERC721, IZNSDomainToken {
      * @notice Burns the token with the specified tokenId
      * @param tokenId The tokenId that the caller wishes to burn/revoke
      */
-    function revoke(uint256 tokenId) external override onlyRegistrar {
+    function revoke(uint256 tokenId) external override onlyRegistrar  {
```
Whytecrowe: extra space

contracts/token/ZNSDomainToken.sol
```diff
@@ -38,11 +38,11 @@ contract ZNSDomainToken is AccessControlled, ERC721, IZNSDomainToken {
...
         _burn(tokenId);
     }
 
-    function setAccessController(address accessController)
+    function setAccessController(address accessController)
```
Whytecrowe: extra space

contracts/upgradeMocks/Mock.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+ /* solhint-disable */
+contract Mock {
```
Whytecrowe: let's maybe make the naming more clear of what this "Mock" is. specify that it's for the upgrade tests

contracts/upgradeMocks/distribution/ZNSEthRegistrarMock.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import { ZNSEthRegistrar } from "../../distribution/ZNSEthRegistrar.sol";
+import { Mock } from "../Mock.sol";
+
+ /* solhint-disable */
+contract ZNSEthRegistrarMock is ZNSEthRegistrar, Mock {}
```
Whytecrowe: same here. let's add "upgrade" to the name or something that clarifies what this is, so that it's clear why these mocks exist. for all the below contracts as well

package.json
```diff
@@ -11,21 +11,22 @@
...
+    "lint": "yarn lint-sol & yarn lint-ts",
     "clean": "hardhat clean",
     "build": "yarn run clean && yarn run compile",
     "typechain": "hardhat typechain",
     "test": "hardhat test",
     "semantic-release": "semantic-release --tag-format='v${version}-dev'",
-    "coverage": "hardhat coverage",
-    "check-coverage": "istanbul check-coverage --statements 90 --branches 87 --functions 89 --lines 90"
+    "check-coverage": "istanbul check-coverage --statements 90 --branches 87 --functions 89 --lines 90",
+    "coverage": "hardhat coverage && yarn check-coverage"
```
Whytecrowe: why did you combine it? did you look into circleci yaml config?
we need these to be different scripts because they are different jobs in the CI. since you've added && here, it will not go through fully if one of the scripts fails, breaking our flow in CircleCI.
we also need to be able to get local coverage files without checking coverage, so that we can actually see what tests we are missing.
let's bring them back to be separate. and before making changes always ensure you are not breaking anything else by doing that.

test/ZNSAddressResolver.test.ts
```diff
@@ -1,195 +1,232 @@
...
   });
 
   it("Should not allow non-owner address to setAddress", async () => {
     await expect(
-      znsAddressResolver.connect(addr1).setAddress(wilderDomainNameHash, addr1.address)
-    ).to.be.revertedWith(
-      "ZNSAddressResolver: Not authorized for this domain"
-    );
+      zns.addressResolver.connect(user).setAddress(wilderDomainHash, user.address)
+    ).to.be.revertedWith("ZNSAddressResolver: Not authorized for this domain");
```
Whytecrowe, line 107: why do you keep reformatting my code for small things like this? how many times will we format back and forth? :)
maybe we should create a guide or something and stick to it, if you are still not satisfied with how I format. Thought we've talked about this already :)
this is all just extra time for small things like this...

test/ZNSAddressResolver.test.ts
```diff
@@ -1,195 +1,232 @@
...
 
     await expect(
-      znsAddressResolver.connect(mockRegistrar)
-        .setAddress(wilderDomainNameHash, addr1.address)
-    )
-      .to.emit(znsAddressResolver, "AddressSet")
-      .withArgs(wilderDomainNameHash, addr1.address);
-  });
+      zns.addressResolver.connect(mockRegistrar).setAddress(wilderDomainHash, hre.ethers.constants.AddressZero)
+    ).to.emit(zns.addressResolver, "AddressSet").withArgs(wilderDomainHash, hre.ethers.constants.AddressZero);
```
Whytecrowe: again here. this looks dirty and hard to read. why did you reformat? how does this help?

test/ZNSDomainToken.test.ts
```diff
@@ -136,13 +148,74 @@ describe("ZNSDomainToken:", () => {
...
+    it("Fails to upgrade if the caller is not authorized", async () => {
+      // UUPS specifies that a call to upgrade must be made through an address that is upgradecall
+      // So use a deployed proxy contract
+      const factory = new ZNSDomainTokenMock__factory(deployer);
+
+      // DomainToken to upgrade to
+      const newDomainToken = await factory.deploy();
+      await newDomainToken.deployed();
+
+      // Confirm the deployer is a governor, as set in `deployZNS` helper
```
Whytecrowe: wrong comment

test/ZNSEthRegistrar.test.ts
```diff
@@ -16,27 +17,26 @@ import { priceConfigDefault } from "./helpers/constants";
...
 require("@nomicfoundation/hardhat-chai-matchers");
 
-
 describe("ZNSEthRegistrar", () => {
   let deployer : SignerWithAddress;
   let user : SignerWithAddress;
   let governor : SignerWithAddress;
   let admin : SignerWithAddress;
-  let randomAcc : SignerWithAddress;
+  let randomUser : SignerWithAddress;
```
Whytecrowe, line 31: :)

test/ZNSEthRegistrar.test.ts
```diff
@@ -527,101 +530,158 @@ describe("ZNSEthRegistrar", () => {
...
+
+    it("Fails to upgrade when an unauthorized users calls", async () => {
+      const registrarFactory = new ZNSEthRegistrar__factory(deployer);
+      const registrar = await registrarFactory.deploy();
+      await registrar.deployed();
+
+      const tx = zns.registrar.connect(randomUser).upgradeTo(registrar.address);
+
+      await expect(tx).to.be.revertedWith(
+        `AccessControl: account ${randomUser.address.toLowerCase()} is missing role ${GOVERNOR_ROLE}`
```
Whytecrowe: use helper here

test/ZNSRegistry.test.ts
Whytecrowe: this was resolved, however nothing has changed

test/helpers/deployZNS.ts
```diff
@@ -61,27 +73,40 @@ export const deployPriceOracle = async ({
...
+    }
+  ) as ZNSPriceOracle;
 
   return priceOracle;
 };
 
 export const deployDomainToken = async (
   deployer : SignerWithAddress,
-  accessControllerAddress : string
+  accessController : string
```
Whytecrowe: let's keep "address" in the name, so we know this is not a contract object. we will call `accessController` the actual contract object. this way we know easily what is needed

test/helpers/deployZNS.ts
```diff
@@ -134,20 +172,14 @@ export const deployZNS = async ({
...
-  zeroVaultAddress ?: string;
-}) : Promise<ZNSContracts> => {
+} : DeployZNSParams) : Promise<ZNSContracts> => {
   const accessController = await deployAccessController({
     deployer,
     governorAddresses: [deployer.address, ...governorAddresses],
     adminAddresses: [deployer.address, ...adminAddresses],
   });
 
+  // TODO AC: Make sure contracts are deployed as proxies and authorize the governor role
```
Whytecrowe: do we still need this comment?

test/helpers/deployZNS.ts
```diff
@@ -195,6 +227,11 @@ export const deployZNS = async ({
     registrar,
   };
 
+  // Final configuration steps
+  // TODO AC: remove all redundant calls here! and delete hashing of the root and the need
+  // for Registrar to be owner/operator of the root
+  await registry.connect(deployer).setOwnerOperator(registrar.address, true);
```
Whytecrowe: why do we need this call here exactly? this seems to go against our AC structure.

test/helpers/types.ts
```diff
@@ -6,11 +6,57 @@ import {
...
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 
 export type Maybe<T> = T | undefined;
 
+export type GetterFunction =
+  (
+    Promise<string> |
+    Promise<boolean> |
+    Promise<BigNumber> |
+    Promise<Array<BigNumber>>
```
Whytecrowe: would this be better if the type parameter is a union type rather than the parent type?

`Promise<string | boolean | ...>` ??

test/helpers/validateUpgrade.ts
```diff
@@ -0,0 +1,57 @@
...
+  ] = await Promise.all(postUpgradeCalls);
+
+  // we're checking previous value after post upgrade calls
+  // to make sure that writing into new storage slots
+  // doesn't overwrite or corrupt the old ones
+  const postVals = await Promise.all(getters);
+
+  preVals.forEach((value, index) => {
+    expect(value).to.eq(postVals[index]);
+  });
```
Whytecrowe, line 51: please, note what I changed here and why

Whytecrowe (changes requested):
Please, address comments.
**Pull from origin before you do this, I've done a couple of commits here.**
Also, please, resolve a whole wagon of compiler warnings that appear when compiling on this branch. They all should be fixed.

contracts/access/AccessControlled.sol
JamesEarle: maybe the change was lost somewhere in git when I merged everything, I didn't manually add this again. Will undo

contracts/distribution/ZNSEthRegistrar.sol
JamesEarle: unintentional

contracts/distribution/ZNSPriceOracle.sol
JamesEarle: removed here

contracts/token/ZNSDomainToken.sol
JamesEarle: Where would you propose it be in this list? As long as they remain in the same order in future upgrades of this contract, it would be fine, no?

test/helpers/deployZNS.ts
JamesEarle: This was in place over a month ago before we had AC and when we still had subdomains. None of the contracts use this and the tests aren't reliant on it anymore so it was just redundant code, I removed it

test/helpers/deployZNS.ts
JamesEarle: nope, removed

test/ZNSDomainToken.test.ts
JamesEarle: fixed for this test

PR #30, opened 2023-05-26
fix: Tenderly DevNets Integration

Whytecrowe:


hardhat.config.ts
```diff
@@ -23,30 +25,46 @@ const config: HardhatUserConfig = {
...
+    username: `${process.env.TENDERLY_ACCOUNT_ID}`,
+  },
 };
 
+// These are needed for tenderly verification to be automatic,
+// they don't seem to work with DevNet for some reason,
+// however Tenderly operations do not work if we don't call
+// this here, it is required for integration with Hardhat.
+// !!! This should be uncommented when using Tenderly DevNet !!!
+// tenderly.setup({ automaticVerifications: true });
```
JamesEarle, line 74: Not sure I fully understand this comment

"These don't work for tenderly DevNet"

but also

"Tenderly operations don't work if we don't include this, uncomment when using Tenderly DevNet"?

src/tenderly/deploy-verify-zns.ts
```diff
@@ -0,0 +1,50 @@
...
+  governorAddresses ?: Array<string>;
+  adminAddresses ?: Array<string>;
+}) => {
+  const zns = await deployZNS({
+    deployer: governor,
+    governorAddresses,
+    adminAddresses,
+  });
+
+  // TODO: find a way to get these names from the object
```
JamesEarle: For a piece of feedback you left in my PR I added the string "ZNSDomainToken" to the constants list for when we deploy a new DomainToken contract. The same could be done here for each contract

src/tenderly/deploy-verify-zns.ts
```diff
@@ -0,0 +1,50 @@
...
+    "ZeroTokenMock",
+    "ZNSAddressResolver",
+    "ZNSPriceOracle",
+    "ZNSTreasury",
+    "ZNSEthRegistrar",
+  ];
+
+  // Verify and add source code to Tenderly
+  // for easy debugging
+  await Object.values(zns).reduce(
```
JamesEarle: Do we need `reduce` here? Seems like it might not be being used how it's intended. A simpler `.forEach` would still do what you're doing here I believe.

src/tenderly/deploy-verify-zns.ts
```diff
@@ -0,0 +1,50 @@
...
+    async (acc, contract, idx) => {
```
JamesEarle: Why is `acc` included? It doesn't seem to get referenced

src/tenderly/devnet-execute.ts
```diff
@@ -0,0 +1,43 @@
...
+ * 2. Launch deploy and operation flow with contracts using Hardhat
+ *
+ * To execute this, uncomment `tenderly.setup()` line in the hardhat.config.ts
+ * then run `yarn devnet` in the terminal.
+ * */
+const execute = async () => {
+  // spawn DevNet on Tenderly with ts-node directly
+  const spawnRes = await execAsync(spawnCommand);
+
+  const opCommand = `${opCommandBase} ${opsPath} ${networkArg}`;
```
JamesEarle, line 29: what's the point of making these 3 different variables when they are all only referenced here, a single time? Just make the command one variable

src/tenderly/run-all-flows.ts
```diff
@@ -0,0 +1,48 @@
+import * as hre from "hardhat";
+import { deployVerifyZNS } from "./deploy-verify-zns";
+import * as ethers from "ethers";
+import { hashDomainLabel } from "../../test/helpers";
+import { BigNumber } from "ethers";
+
+
+const domainName = "wilder";
+const domainHash = hashDomainLabel(domainName);
+const tokenId = BigNumber.from(domainHash);
```
JamesEarle, line 9: If the `tokenId` internally is the `uint256(domainHash)` then won't this not work? This is just going to be the BigNumber of that `bytes32` object?

src/tenderly/deploy-verify-zns.ts
JamesEarle, line 1: Consider camelCase for file names with `.ts` extensions

src/tenderly/spawn-devnet.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const {
+  TENDERLY_ACCESS_KEY,
+  TENDERLY_PROJECT_SLUG,
+  TENDERLY_ACCOUNT_ID,
+  TENDERLY_DEVNET_TEMPLATE,
+} = process.env;
+
+// eslint-disable-next-line max-len
+const command = `tenderly devnet spawn-rpc --project ${TENDERLY_PROJECT_SLUG} --template ${TENDERLY_DEVNET_TEMPLATE} --account ${TENDERLY_ACCOUNT_ID}  --access_key ${TENDERLY_ACCESS_KEY}`;
```
JamesEarle, line 19: You can use `\` to have multiline commands in code instead of disabling eslint

``

src/tenderly/spawn-devnet.ts
```diff
@@ -0,0 +1,44 @@
...
+
+
+const spawnDevNet = async () => {
+  const { stderr } = await asyncExec(command);
+  const devNetUrl = stderr.trim().toString();
```
JamesEarle: This feels super weird... we're intentionally running a command that fails, then building a variable from stderr?

src/tenderly/spawn-devnet.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  console.log(`DEVNET_RPC_URL=${ devNetUrl }`);
+
+  // if file doesn't exist, create it
+  if (!fs.existsSync(".env")) {
```
JamesEarle, line 30: just add this to the options for the `writeFileSync` call below, it will append to an existing file or create a new file if one doesn't exist automatically using the `w` flag 

`fs.writeFileSync(".env", { flag : "w" });`

src/tenderly/spawn-devnet.ts
```diff
@@ -0,0 +1,44 @@
...
+    fs.writeFileSync(".env", "");
+  }
+  const fileContent = fs.readFileSync(".env", "utf8");
+
+  const newFileContent = fileContent.replace(/DEVNET_RPC_URL=.*/g, "");
+  fs.writeFileSync(".env", newFileContent);
+  fs.appendFileSync(".env", `DEVNET_RPC_URL=${ devNetUrl }`);
```
JamesEarle, line 37: Can you explain this flow? 

We replace `DEVNET_RPC_URL=` in the file with an empty string`""`,
Then we write that empty string to the file
Then we append to that file with `DEVNET_RPC_URL=${devNetUrl}`

Why do we do it this way? Would not be simpler to see if `.env` contained a devNetUrl already and if so, replace it, or if it doesn't add it? This is presuming the URL is different each time as well.

src/tenderly/deploy-verify-zns.ts
Whytecrowe: this is kind of what I did here, only these are not constants yet. I don't like this approach though.
it is weird to me that I can't seem to find a contract name in the contract object. No idea why they didn't include it. or maybe I'm just blind...

src/tenderly/deploy-verify-zns.ts
Whytecrowe: `forEach` may not work with promises properly as some other looper functions. or you have to trick and use `Promise.all()` making it harder to read. here it's simple await wherever you want.
the reason I'm using `reduce` here is because it can properly await promises at correct times.
I also really like `reduce` because it can be used in whichever way you want and it works a bit differently than other array methods. with async code iteration through reduce is my favorite because it's predictable.

reduce is very popular for async code. you don't have to accumulate necessarily if you don't need to, but the method itself is very good.

src/tenderly/deploy-verify-zns.ts
Whytecrowe: because you need to use other params that are after it.

src/tenderly/devnet-execute.ts
Whytecrowe: it's more future proofing. so we can change parts of it independently if path or network changes. but I will look into it. maybe it's not necessary

src/tenderly/run-all-flows.ts
Whytecrowe: not sure what you mean here. bignumber is casting number into a bytes hash. when we use it as a number it converts that to an int. why would this not work?

src/tenderly/deploy-verify-zns.ts
Whytecrowe: why camelCase? most of the file "conventions" are usually kebab-case, no? never worked with camelCase files before. feels kinda strange...

src/tenderly/spawn-devnet.ts
Whytecrowe: I don't think it fails necessarily. They probably just use this channel to pass the data on their side. not sure why. if it failed, it won't spawn the devnet.... or shouldn't...

src/tenderly/spawn-devnet.ts
Whytecrowe: actually, I think "w" is already there by default. will check if it works

src/tenderly/spawn-devnet.ts
Whytecrowe: I actually lifted this piece of code :))
The way I see it, it's to also make sure the old link is removed before the new one added. Maybe for potential conflicts or to not fuse 2 values together. Not sure this can even happen, though...

"to see if .env contained a devNetUrl already and if so, replace it, or if it doesn't add it" - isn't that what is being done here though?

src/tenderly/deploy-verify-zns.ts
JamesEarle: Seems more consistent to the other files in the repo but not a big deal

src/tenderly/spawn-devnet.ts
JamesEarle: I guess it didn't logically fail, but if it's coming through stderr it's failing to the system, but if that's how they've done it you have to read it this way. Just a weird choice on their part....

src/tenderly/spawn-devnet.ts
JamesEarle: I guess that is what it's doing, but it feels like you can just use a flag in `fs.writeFileSync` to overwrite contents of a file and then you can avoid doing this sort of round about check

https://nodejs.org/api/fs.html#file-system-flags

src/tenderly/spawn-devnet.ts
JamesEarle: Please confirm, I've seen a few posts on stackoverflow that it is the default according to the docs but it doesn't behave that way

src/tenderly/run-all-flows.ts
JamesEarle: Did you confirm this works? It wouldn't be an `int` yet explicitly as you haven't called to modify it and it would just be "0x..." internally still. I guess it just does this implicitly in the function call below?

src/tenderly/spawn-devnet.ts
Whytecrowe: agreed

PR #32, opened 2023-06-12
fix: Publish zToken contract as an npm package and import into zNS

JamesEarle:


test/helpers/deployZNS.ts
```diff
@@ -207,7 +243,7 @@ export const deployZNS = async ({
     deployer,
     accessController.address,
     priceOracle.address,
-    zeroTokenMock.address,
```
Whytecrowe: it's still a mock. not actual zero token. let's specify this, so nobody deploys it by mistake to a network

package.json
```diff
@@ -43,9 +43,10 @@
     "@types/mocha": "9.1.0",
     "@types/node": "^18.15.11",
     "@zero-tech/eslint-config-cpt": "0.2.7",
+    "@zero-tech/ztoken": "^1.0.1",
```
Whytecrowe: let's make a specific version here rather than caret

Whytecrowe (changes requested):
couple of minor things

test/helpers/deployZNS.ts
JamesEarle: Makes sense, I'll add this as well as a comment specifying that this script is used for testing only

Whytecrowe (approved):

PR #36, opened 2023-06-28
feat: [PROTO] Zero Deploy Campaign [WIP]

Whytecrowe:


src/deploy/campaign/deploy-campaign.ts
```diff
@@ -0,0 +1,85 @@
...
+  dbAdapter : BaseStorageAdapter;
+  logger : Console;
+  config : IDeployCampaignConfig;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [name : string | symbol] : any;
+
+  private static indexedHandler : ProxyHandler<DeployCampaign> = {
+    get: (target, prop) => {
+      if (typeof prop === "string" && !!target.state.contracts[prop]) {
```
JamesEarle: I believe in this case there's no need to use the double not `!!`, as it would be checked for its truth value anyways in the `if`. Same for some of the other uses in this file

src/deploy/campaign/deploy-campaign.ts
```diff
@@ -0,0 +1,85 @@
...
+    },
+  };
+
+  constructor ({
+    missions,
+    deployer,
+    dbAdapter,
+    logger,
+    config,
+  } : ICampaignArgs) {
```
JamesEarle, line 41: If you've already made a type `ICampaignArgs` for the code, consider making the input object just a single variable

src/deploy/campaign/deploy-campaign.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.deployer = deployer;
+    this.dbAdapter = dbAdapter;
+    this.logger = logger;
+    this.config = config;
+
+    const campaignProxy = new Proxy(this, DeployCampaign.indexedHandler);
+
+    // instantiate all missions
+    this.state.instances = missions.map(
+      (mission : TDeployMissionCtor) => new mission({
```
JamesEarle, line 57: Why are we passing this var? Can we not just call a new object directly from the imported type?

src/deploy/campaign/deploy-campaign.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    this.logger.debug("DeployCampaign initialized.");
+
+    return campaignProxy;
+  }
+
+  async execute () {
+    return this.state.instances.reduce(
+      async (
+        acc : Promise<void>,
```
JamesEarle, line 74: what is `acc` meant to be here?

src/deploy/constants.ts
```diff
@@ -0,0 +1,8 @@
+
+export const ProxyKinds = {
```
JamesEarle, line 3: This is just an object with three mutually exclusive properties. Make this a struct instead

src/deploy/deployer/deployer.ts
```diff
@@ -0,0 +1,48 @@
...
+  constructor () {
+    this.hre = hre;
+  }
+
+  async deployProxy ({
+    contractName,
+    args,
+    kind,
+  } : {
+    contractName : string;
```
JamesEarle: Create a type for this like you have in other code

src/deploy/missions/base-deploy-mission.ts
```diff
@@ -0,0 +1,127 @@
...
+  async needsPostDeploy () {
+    return Promise.resolve(true);
+  }
+
+  async postDeploy () {
+    return Promise.resolve();
+  }
+
+  async execute () {
+    if (await this.needsDeploy()) {
```
JamesEarle, line 117: does this need to be an async function? could it not be a boolean prop?

src/deploy/campaign/deploy-campaign.ts
Whytecrowe: this is more thorough. converting the value into a boolean. this is never bad usually, while just relying for it to check the value presence might not always work, in theory

src/deploy/campaign/deploy-campaign.ts
Whytecrowe: no need to convert it to a single variable and then destructure or access again from it in the constructor. changing one destructuring for another here

src/deploy/campaign/deploy-campaign.ts
Whytecrowe: not sure what you mean

src/deploy/missions/base-deploy-mission.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  deployArgs () : TDeployArgs {
+    return [];
+  }
+
+  getArtifact () {
+    return this.campaign.deployer.getContractArtifact(this.contractName);
+  }
+
+  buildDbObject (hhContract : Contract, implAddress : string | null) : IContractDbObject {
```
JamesEarle, line 74: Consider making `type Maybe<T> = T | undefined` for uses like with implAddress

Follow up question, does this value have to be `null` for a specific reason?

src/deploy/campaign/deploy-campaign.ts
Whytecrowe: it needs to be there so we can get the next param, which is `mission`. this is how reduce works or using parameters. :) we also need to await it, because it should automatically return acc each iteration. it is just a placeholder to make the method work correctly

src/deploy/constants.ts
Whytecrowe: what struct?
this is a map of constant variables to make sure we use the same strings everywhere. you can call it an NodeJS version of enum.

src/deploy/deployer/deployer.ts
Whytecrowe: I will

src/deploy/missions/base-deploy-mission.ts
Whytecrowe: we have async calls inside of this hook often, so, yes, it does.

src/deploy/missions/base-deploy-mission.ts
Whytecrowe: `null` is specifically here instead of `undefined`.
undefined type is too general and can be that because of ton of different things that can happen in code.
`null` is a more specific type here because it signifies that dbAdapter did not find anything.

src/deploy/campaign/deploy-campaign.ts
JamesEarle: If the value isn't present it will just be `undefined` and so will resolve to `false`

While it doesn't do any harm to have the additional `!!` it doesn't make a difference if it's in an. `if(variable)` and `if(!!variable)` would be equivalent whereas `const x = variable` wouldn't be equivalent with `const x = !!variable`

PR #41, opened 2023-07-26
Subdomain Subsystem Prototype

Whytecrowe:


contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+        // TODO sub: can this be combined with the same check in the Main Registrar ??
+        require(
+            rootRegistrar.isOwnerOf(domainHash, msg.sender, IZNSRegistrar.OwnerOf.BOTH),
+            "ZNSSubdomainRegistrar: Not the owner of both Name and Token"
+        );
+
+        rootRegistrar.coreRevoke(domainHash);
+
+        // TODO sub: do we store these as addresses or interfaces in the struct ??
+        address paymentContract = address(distrConfigs[parentHash].paymentContract);
```
durienb, line 120: Could do an erc165 interface check to be safe

contracts/distribution/IZNSRegistrar.sol
```diff
@@ -1,15 +1,24 @@
...
+import { IDistributionConfig } from "./subdomains/IDistributionConfig.sol";
 
-interface IZNSRegistrar {
+
+interface IZNSRegistrar is IDistributionConfig {
+
+    enum OwnerOf {
+        NAME,
+        TOKEN,
+        BOTH
```
durienb, line 12: So is it possible to have a different owner for the token than the domain?
If so what is the benefit of that?
I would generally think this should be avoided and just make the token implicitly owned by the domain holder.

contracts/distribution/ZNSRegistrar.sol
```diff
@@ -143,13 +185,23 @@ contract ZNSRegistrar is AccessControlled, UUPSUpgradeable, IZNSRegistrar {
     function revokeDomain(bytes32 domainHash)
     external
     override
-    onlyNameOwner(domainHash)
-    onlyTokenOwner(domainHash)
     {
+        require(
+            isOwnerOf(domainHash, msg.sender, OwnerOf.BOTH),
+            "ZNSRegistrar: Not the owner of both Name and Token"
+        );
```
durienb, line 192: BOTH used here instead of the two checks onlyNameOwner and onlyTokenOwner but I'm not sure this is saving anything. The isOwnerOf function has an if elseif tree that could just be avoided. 
I would generally prefer the two checks you had before, and hopefully eliminate the extra storage slot for BOTH.
But haven't run it myself, is there a real savings here?

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    function registerSubdomain(
+        bytes32 parentHash,
+        string calldata label,
+        address domainAddress,
+        DistributionConfig calldata distrConfig
+    ) external override returns (bytes32) {
+        // TODO sub: make the order of ops better
+        DistributionConfig memory parentConfig = distrConfigs[parentHash];
```
durienb, line 53: Should just use distrConfigs[parentHash] everywhere here instead of setting a new local variable. The compiler automatically optimizes any values that are always the same (same outputs for same inputs). 
Setting a local variable like this wastes gas. 
You'd think that the compiler would take care of optimizing out the cost of setting this local variable, but it doesn't because of how storage works.
This is just a solc quirk, any other language you'd do it this way.

durienb: There are times when setting a local variable is a savings, it depends on the evm calls happening and how static the function is. So maybe this is saving but don't think so.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+
+        bytes32 subdomainHash = hashWithParent(parentHash, label);
+
+        require(
+            !registry.exists(subdomainHash),
+            "ZNSSubdomainRegistrar: Subdomain already exists"
+        );
+
+        uint256 price;
+        uint256 fee;
```
durienb, line 75: Could also avoid setting these variables if you just read the getPriceAndFee directly into the config values. But would have to rework the structure here and maybe add getPrice and getFee separately

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
durienb: Also on this line you're using distrConfigs[parentHash] instead of the parentConfig. But this is probably how they should all be.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+            subdomainHash,
+            label,
+            msg.sender,
+            domainAddress
+        );
+
+        if (address(distrConfig.pricingContract) != address(0)
+            && address(distrConfig.paymentContract) != address(0)) {
+            setDistributionConfigForDomain(subdomainHash, distrConfig);
+        }
```
durienb, line 105: It seems there's some potential for misconfigured distrConfig, but i'm not really sure what checks are important because if you're having to check everything here it might be too inefficient.

contracts/distribution/subdomains/payments/ZNSDirectPayment.sol
```diff
@@ -0,0 +1,87 @@
...
+        // setting paymentToken to 0x0 address means free domains
+        // to save on tx costs, we avoid transfering 0
+        // TODO sub: do we need an event here that will signify it was a free payment ??
+        if (address(config.paymentToken) == address(0)) return;
+
+        config.paymentToken.safeTransferFrom(
+            payer,
+            config.beneficiary,
+            amount + fee
+        );
```
durienb, line 39: It's unfortunate that this will require an approval tx beforehand.
This might become a bigger problem too if approval needing to be given to multiple different contracts in the flow.

We might consider making the switch to erc777 tokens, where we have an onReceived hook that could run this logic on a token transfer, eliminating the approval tx.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    function revokeSubdomain(bytes32 parentHash, bytes32 domainHash) external override {
+        // TODO sub: can this be combined with the same check in the Main Registrar ??
+        require(
+            rootRegistrar.isOwnerOf(domainHash, msg.sender, IZNSRegistrar.OwnerOf.BOTH),
+            "ZNSSubdomainRegistrar: Not the owner of both Name and Token"
+        );
+
+        rootRegistrar.coreRevoke(domainHash);
```
durienb, line 117: Concerned about the order of operations here. 
It's hard to see through all the consequences exactly of doing the coreRevoke here, but it seems there's some refund, state setting logic in this coreRevoke call that might cause the code below to fail.
Something that tests should at least be checking for if they aren't already.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    function getAccessController() external view override(AAccessControlled, IZNSSubdomainRegistrar) returns (address) {
+        return address(accessController);
+    }
+
+    function setAccessController(address accessController_)
+    external
+    override(AAccessControlled, IZNSSubdomainRegistrar)
+    onlyAdmin {
```
durienb, line 226: Because the admin can do this and other things here like the setRootRegistrar, the admins will always be able to do things like brick domains. Has a significant centralizing effect, we should at least make sure that we have a colective understanding and are ok with the admin powers that we have and dont have.

durienb: I generally think more power to domain owners and less to admins is good but it's a hard thing to balance in a complex system.

contracts/distribution/subdomains/abstractions/AZNSPayment.sol
```diff
@@ -0,0 +1,40 @@
...
+        bytes32 indexed domainHash,
+        address indexed payer,
+        uint256 amount,
+        uint256 fee
+    );
+
+    struct PaymentConfig {
+        IERC20 paymentToken;
+        address beneficiary;
+    }
```
durienb, line 23: Might be able to simplify this whole thing by just having a strict payment token for all domains like WILD or ZERO

contracts/distribution/subdomains/abstractions/AZNSPayment.sol
```diff
@@ -0,0 +1,40 @@
...
+        bytes32 parentHash,
+        bytes32 domainHash,
+        address payer,
+        uint256 amount,
+        uint256 fee
+    ) external virtual;
+
+    function refundsOnRevoke() external pure virtual returns (bool) {
+        return false;
+    }
```
durienb, line 39: These refundsOnRevoke functions defining the behavior of the contracts are possibly an issue because there's no way to verify logic, like in an erc-165 check you can only verify the implementation. So there isn't a good way to make sure that these functions are returning the appropriate true/false state.

And just in general seeing this tells me that the logic should probably be done elsewhere, like here?, so that this isn't needed.

contracts/distribution/subdomains/abstractions/AZNSPricing.sol
```diff
@@ -0,0 +1,17 @@
...
+    // TODO sub: do we leave only return of the price
+    //  or do we leave fee values here as well?
+    function getPrice(
+        bytes32 parentHash,
+        string calldata label
+    ) external virtual view returns (uint256);
+
+    function feeEnforced() external pure virtual returns (bool) {
+        return false;
+    }
```
durienb, line 16: Same thing with this as in AZNSPayment. Probably not good to have the functionality of contracts ultimately external to the contract.

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
```diff
@@ -0,0 +1,208 @@
...
+    function setAccessController(address accessController_)
+    external
+    override
+    onlyAdmin {
+        _setAccessController(accessController_);
+    }
+
+    function getAccessController() external view override returns (address) {
+        return address(accessController);
+    }
```
durienb, line 174: There's a lot of repetition of these getters and setters for the contracts that get attached like the accessController.
It seems like these should be consolidated if possible.

contracts/distribution/subdomains/pricing/ZNSFixedPricing.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import { AZNSPricing } from "../abstractions/AZNSPricing.sol";
+import { IZNSRegistry } from "../../../registry/IZNSRegistry.sol";
+import { AAccessControlled } from "../../../access/AAccessControlled.sol";
+import { ARegistryWired } from "../../../abstractions/ARegistryWired.sol";
+
+
+contract ZNSFixedPricing is AAccessControlled, ARegistryWired, AZNSPricing {
```
durienb, line 10: Again should probably consider consolidating all these different pricing models if possible.
Or just deciding on one and simplifying all this down.
These are maybe making the system more feature complete but it's probably at the expense of a significant complication to devs and users.

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
```diff
@@ -0,0 +1,208 @@
...
+        bytes32 indexed domainHash,
+        uint256 maxPrice,
+        uint256 minPrice,
+        uint256 maxLength,
+        uint256 baseLength,
+        uint256 precisionMultiplier,
+        uint256 feePercentage
+    );
+
+    uint256 public constant PERCENTAGE_BASIS = 10000;
```
durienb, line 35: Might want this to be configurable too

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
```diff
@@ -0,0 +1,208 @@
...
+    constructor(address _accessController, address _registry) {
+        _setAccessController(_accessController);
+        setRegistry(_registry);
+    }
+
+    function getPrice(
+        bytes32 parentHash,
+        string calldata label
+    ) public view override returns (uint256) {
+        uint256 length = label.strlen();
```
durienb, line 49: Don't set this local variable just use label.strlen() on 51

contracts/distribution/IZNSRegistrar.sol
Whytecrowe: This is actually taken from ENS. The idea is that owner is checked for many cases and if you want to make a contract that would handle many ops, this contract has to be the owner of the Name. You still keep the token on your local wallet, then if you want to switch back cause you destroyed the contract or want to revoke, you can call `reclaim()` as a Token owner to become the owner of the Name in Registry again, claiming full ownership and blocking the contract from doing any more ops. You can also just assign an `operator` instead, but we are still figuring out on the difference in access between `owner` and `operator`.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: what do you mean by `parentConfig`?

contracts/distribution/ZNSRegistrar.sol
Whytecrowe: with the if-else we are only running the code inside the block that matches instead of running both checks for every case.
since this function is public, it can be called from any contract to validate the ownership quicker rather than creating checks for every contract. instead, we use Registrar as the checker.

Whytecrowe: what storage slot for BOTH are you talking about? this is a memory var.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: could you please elaborate?
I don't think solc optimizes anything here. Usually every time you read from state var you pay an SLOAD, even if you are reading the same var, no?
we are reading this struct 6 times here. if I do not use memory var, I am paying for 6 SLOADs, instead of 1 SLOAD, 1 MSTORE and 5 MLOADs (which is way cheaper). did I miss some kind of new optimization of solc?

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: I was thinking about it, and still working on how to structure these abstracts in the best way.
However, if I split it in 2 functions, we will need to do 2 calls for a fee based pricing, meaning we will pay double the call gas here instead of calling one function to get both. external call here wouldn't be too much gas, but I am not sure the code appearance here is more important than gas

Whytecrowe: I also use these vars twice here (added in another PR), so making 2 calls instead of a single call and 2 reads seems better to me.

contracts/distribution/subdomains/abstractions/AZNSPricing.sol
Whytecrowe: yeah, still thinking about the structure of these abstracts. I added this for now since user developed modules can override these to not behave properly, this way we ensure that a pricing with fee will always return true here.

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
Whytecrowe: still thinking about it. the problem comes from proxies. I could add these to the AAccessControlled, but it would not let us call if from the interface variable contract in external contract. so I added the virtual functions to the abstract to make sure they are always implemented. any ideas on this?

contracts/distribution/subdomains/pricing/ZNSFixedPricing.sol
Whytecrowe: Not sure what you exactly mean here. Consolidate, meaning, put all options in one contract? That would defeat the purpose of "modular" system, where we can evolve each module separately and let users create their own modules.
By "consolidating" we are losing the modularity and ease of extensibility. We are overcrowding one contract with too much logic, locking the potential updates under the size limit.

I am open to hear the opposite argument though, if you have a good idea of how this can be improved and why.
also, which complications exactly you are talking about?

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
Whytecrowe: why exactly? to go beyond 2 decimals in a percentage value? would that be needed you think?

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
Whytecrowe: good catch, but I believe, the wrong one. later in the code we call the strlen() again, which is an error. we can just use this var, since it's used twice. that's better than calling the lib twice, I believe.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
durienb: It should optimize them. Something to test maybe, have seen tests of this show variance in gas cost vs. bytecode size in other systems. 
Since we have the tenderly stuff in it might be easier to test?

contracts/distribution/subdomains/pricing/ZNSFixedPricing.sol
durienb: Was trying to say that since the goal is modularity and flexibility, it might have been sufficient to just abstract all this and not really define any of these options it except for the base case. 
And options we release ourselves should probably be maintained/monitored.

But yeah this is good, it's feature complete and there are different config options. There's probably more optional cases we'll want or need to provide at some point anyway.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+            price = parentConfig.pricingContract.getPrice(parentHash, label);
+        }
+
+        parentConfig.paymentContract.processPayment(
+            parentHash,
+            subdomainHash,
+            msg.sender,
+            price,
+            fee
+        );
```
durienb, line 92: So if you hit the else on 82, then you've created a local variable for the fee but didn't ever set it and just passed it into processPayment as 0.

The fee logic could be on getPrice(), with the processPayment deciding who to pay with what it's given?

What's the reason they're separate? maybe combine the two contracts?

This is related to the problem below where the feeEnforced() is defined and there's the functions that just return a bool

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
```diff
@@ -0,0 +1,229 @@
...
+        address registrant,
+        bool allowed
+        // TODO sub: should we allow Registrar role to call this ??
+        //  if so - why ?? what can we call from there ??
+        //  should we cut Registrar out of this ??
+    ) external override onlyOwnerOperatorOrRegistrar(domainHash) {
+        distributionWhitelist[domainHash][registrant] = allowed;
+
+        emit WhitelistUpdated(domainHash, registrant, allowed);
+    }
```
durienb, line 206: We could have a similar function but that whitelists a range of token ids and a contract address, to whitelist NFT holders.

contracts/distribution/subdomains/abstractions/AZNSPricing.sol
durienb: This is related to the issue above, where you're using this feeEnforced() check and then running price logic based on it.

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
durienb: yes, its not being used on 53
you should just delete 49 and use label.strlen() on 51 too though

worth a sanity test at least.

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
durienb: Yeah I see why it's this way, as these are the modular options.

I think the question is just what kinds of things are defined at the base level of the registrar and what's in the modular attachements.

Any time something is repeated a lot though, I think it's just evidence we probably could stick to some base case at the higher level.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: This has been a thing in Solidity since forever and is one of the main advices on how to reduce gas costs - using memory vars for multiple state reads. But I can test it, just to be sure.

Whytecrowe: This is interesting. I must've missed this update. The memory var usage was always the way to go.
It still seems like the way to go, but I was surprised by the result of testing.
I did 2 ways of testing: estimateGas from ethers.js and running it on tenderly evm. The results are different and I'm not sure why thh diff is so big between those 2 methods.

Results:
memory var: 222002 gas - ethers (336,855 gas - tenderly)
direct state reads: 222427 gas - ethers (337,276 gas - tenderly)

We still see that using the memory var reduces the gas by around 400gas. Which is a very small amount. And this is strange.
Could you elaborate (if you know), how exactly these state reads are optimized now?

Whytecrowe: To read later:
This has been discussed and researched.
"Each storage read (opcode sload) of the same slot costs 2,100 gas the first time during a transaction, and then 100 gas each other time during the same transaction. (After the EIP-2929 implemented in the Berlin hardfork. Before that, it was 800 for each read no matter how many times you performed the reads.)" (c)

What Damien also meant in above comments is that creating a memory var can reduce stack size (correct me if I'm wrong). Want to specify here that memory and stack are different things. We often see an error like "stack too deep, try reducing local variables". The limitation comes not from the stack size (which is 1024 elements), but from some opcodes being able to handle up to 16 elements (SWAP, DUP, etc.). This creates not a problem of stack being too populated, but the way local memory vars are structured and used within a *single function*. Meaning, a single function can only hold up to 16 variables.
While I am still not sure how much of the stack persists between functions, 16 vars in a total callstack is not a limit anymore, since stack (if not persisted) would be swapped to put newer vars on top. Since memory does not persist between function calls, the 16 var limit is not the problem anymore. It is the availability of a variable in the stack (being in the top 16 elements).

So all that is happening here is using memory to reduce state reads (SLOAD - 2100/100 gas) with memory reads (MLOAD - 3 gas). @durienb claims that in this case we do not use SLOAD 6 times, but only use once and then use it from the stack. I am not sure this is true, couldn't verify yet.
If we use approx calculation in gas diff, we get:
Only state reads: 2100 + 100 * 5 = 2600 gas (for 6 total reads)
State read with memory var: 2100 + 3 * 6 = 2118 gas (with one MSTORE)
26000 - 2118 = 482 gas - which is suspiciously close to 500 gas for 5 extra SLOADs, but this is not confirmed.

Conslusion:
Since we do not have a stack problem or too many vars in this function, we can keep the memory var and save about 500 gas here. The other option can be done, but deemed unnecessary for now.
500 gas was not worth this time and discussion, but the resulted research and new knowledge about stack and memory connections and some opcodes was definitely beneficial as a result. :)

contracts/distribution/subdomains/pricing/ZNSFixedPricing.sol
Whytecrowe: I think they are different enough in the state structure they use vs the process with which they work. we do have some abstracted, and more work will be done to make it better. but I wanted to add full functionality first in order to make the best decision on what to abstract.

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
Whytecrowe: you really hate local vars, huh? :)) even two identical calls is better? :)

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: The goal was to make it work with whatever Payment contract. The payment contract then decides what to do with the fee based on it being 0 or not.
If we combine these two contracts, then the modularity and possibility to combine different pricing with different payments goes away. Passing a 0 value in some cases seems like a reasonable sacrifice for the ability to configure.
But I will look into this more, since there might be a way to make it better. Fees are still kind of janky and not finalized. Just wanted to make it work with whatever contract as a start.
Also, `feeEnforced()` call is on the Pricing contract, not payment one. And might be eliminated altogether soon. One of the next PRs after this already remove non-fee pricing contracts.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: Misconfigured where and how exactly?
This is basically an option for the user to either set the config upon registration or set it later.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: `coreRevoke()` wipes the main domain data only, which is in registry and domain token. It also now sets whatever price was set by the revokee to 0, so that new registrations do not have to pay to the guy who has no domain anymore.
refunds so far are only issued by StakePayment, which keeps it's own state of how much each person staked, so that wiping the price doesn't do anything.

Whytecrowe: and yes, there's already a ton of tests for refunds :)

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: yes! that's a great point! need to add this for new registrants.
Not sure, though, what you mean by tokenIds and a contract address? which NFT holders?

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
Whytecrowe: I do agree partially. We can re-set access controller or any other contract connection to a different one. I can't see how we can brick existing domains with it though. Registry + DomainToken will keep all the data anyway, and those are 2 main points that prove your domain exists. even if Registrar is broken, all existing domains should still be accessible. maybe not reclaim or revoke though if Registrar is down or changed.

I can't see how we can give access to these functions to users. We can't and shouldn't.

I do agree with the fact that we need to understand, we have centralization here. But that was chosen from the get-go. Admins can in the future be DAO contracts or participants. Might make it more decentralized.
I don't really know yet how we can avoid this. It's apparent that contract connections have to be set by specific people, otherwise system will break on the first day :) as you said, complexity makes it hard to avoid

contracts/distribution/subdomains/abstractions/AZNSPayment.sol
Whytecrowe: yes, but that goes against core requirements :)

contracts/distribution/subdomains/abstractions/AZNSPricing.sol
Whytecrowe: still not sure what the problem is exactly.

contracts/distribution/subdomains/payments/ZNSDirectPayment.sol
Whytecrowe: I mean, yes... But also, every single payment logic on any contract ever requires an approval tx. The reason for it existing is to give you control over fund movements from your wallet. While it is inconvenient and requires extra steps, the benefit of it is obvious, and most people who use eth sc protocols are used to it.
As an alternative we can utilize multicall contracts to pack those 2 txes together (approve + register).
ERC-777 is still susceptible to many attacks, is hard to implement and maintain. at least for now.
Also, this flow is for user chosen tokens. If we only support ERC-777, we will support a very tiny amount of tokens, non of which are popular in ETH ecosystem (yet).

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
durienb: "What Damien also meant in above comments is that creating a memory var can reduce stack size"

no I meant the opposite, it increases it. it uses up an extra one of the 16 mem vars you have.

even though it's not a big problem here, it was in the old implementation, which I solved by removing local variables, so I think it makes sense to at least be aware of this here and not run into the same issues as before.

--

from what I understand what you said about your tests is that when you don't declare the variable, you saved 18 gas (negligible) and saved the use of the extra storage slot.
so this thing about the extra SLOADs isn't quite correct or there's something else to it, as the tests show that contrary to saving 500 gas this way it actually is basically the same amount of gas, but no extra storage slot used.

so given all that, it doesn't really make sense to declare a local variable when you don't need to.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
durienb: As in, the contracts provided don't adhere to the erc165 interface.
So in the places here where you're calling things like parentConfig.accessType, these will break if the contract doesn't have this member.

But it may just be fine, as you've said it's up to the module to work properly.

contracts/distribution/subdomains/ZNSSubdomainRegistrar.sol
durienb: The NFT holders that own token IDs at a contract address :)

contracts/distribution/subdomains/abstractions/AZNSPricing.sol
durienb: It's just that the price logic is separate from getPrice.

It's possible for this AZNSPricing contract to intend to enforce a fee because feeEnforced() returns true, but for ultimately the fee to not get paid because the middleman contract where the fee is then passed into the payment contract isn't doing that correctly for some reason, which isn't happening but just an example of a problem that can arise from the fee logic ultimately being external to the pricing contract like this.

contracts/distribution/subdomains/payments/ZNSDirectPayment.sol
durienb: Well I more meant with our own tokens.
And the point being it has an onReceived() function just like you have in ERC712 on NFTs.

I don't see much benefit to the approval pattern, extra tx is generally bad for the ux I think.
Also infinite approvals tend to cause issues.

Do we have an idea of every contract that will have to be approved, is it just here?

contracts/distribution/subdomains/pricing/ZNSAsymptoticPricing.sol
durienb: It depends on what the call is so usually just test it if it's in question.
I understand you did some testing of it so all good.

PR #59, opened 2023-10-24
fix: [audit] ZNS-6 - Get a Free Domain with Undefined Price Configuration 

Whytecrowe:


test/helpers/register-setup.ts
```diff
@@ -165,19 +167,25 @@ export const registrationWithSetup = async ({
...
-      fullConfig.priceConfig as ICurvePriceConfig,
+      {
+        ...fullConfig.priceConfig as ICurvePriceConfig,
+        isSet: true,
+      },
     );
   }
 
-  if (fullConfig.paymentConfig.beneficiary !== AddressZero) {
+  if (fullConfig.paymentConfig.beneficiary !== AddressZero && setConfigs) {
```
JamesEarle, line 188: mayeb simplify this and the above `if` statements that all rely on `setConfigs` by having one `if(!setConfigs) throw ...` at the top

PR #60, opened 2023-10-25
fix: [audit] ZNS-8 Mintlist

Whytecrowe:


JamesEarle (approved):
nested mapping go brrrr :shipit:

PR #61, opened 2023-10-25
fix: [audit] ZNS-1, ZNS-2, vulnerability due to lack of string validation

JamesEarle:
String validation now exists to only allow domain names to be alphanumeric characters that match `[a-z0-9-]`

By adding this, creating a single domain or subdomain with the name `0://wilder.2ld.3ld.4ld` is not possible because the `.` character does not match this pattern. Any domain name that includes a `.`, a capital letter, or any special character like `!()@#$%^&*?` is not valid. We also no longer allow unicode characters for special things like emojis or letters with accent marks.

Validation function exists in `StringUtils` library

This addresses ZNS-1 in a way as well. Spoofing is still possible because people can mint domains like `w1lder` and `wild3r` which can't really stopped entirely unless we also disallow numbers. This change does, however, address capital letters and other possibly abusable special characters so domains like `Wilder` can't be created

### DIFF
--- a/contracts/price/ZNSCurvePricer.sol
+++ b/contracts/price/ZNSCurvePricer.sol
@@ -61,6 +61,9 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         bytes32 parentHash,
         string calldata label
     ) public view override returns (uint256) {
+        // Confirms string values are only [a-z0-9]
+        label.validate();
+
         uint256 length = label.strlen();
         // No pricing is set for 0 length domains
         if (length == 0) return 0;
--- a/contracts/price/ZNSCurvePricer.sol
+++ b/contracts/price/ZNSCurvePricer.sol
@@ -298,8 +298,6 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
             / config.precisionMultiplier * config.precisionMultiplier;
 
         return price;
-        // To avoid an issue where the config might return less than the min price at a point,
-        // if (price < config.minPrice) return config.minPrice;
     }
 
     /**
--- a/contracts/price/ZNSCurvePricer.sol
+++ b/contracts/price/ZNSCurvePricer.sol
@@ -294,10 +294,8 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         if (length <= config.baseLength) return config.maxPrice;
         if (length > config.maxLength) return config.minPrice;
 
-        uint256 price = (config.baseLength * config.maxPrice / length) 
+        return (config.baseLength * config.maxPrice / length)
             / config.precisionMultiplier * config.precisionMultiplier;
-
-        return price;
     }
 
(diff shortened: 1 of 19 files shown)


contracts/price/ZNSCurvePricer.sol
```diff
@@ -291,9 +294,10 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         if (length <= config.baseLength) return config.maxPrice;
         if (length > config.maxLength) return config.minPrice;
 
-        return
-        (config.baseLength * config.maxPrice / length)
-        / config.precisionMultiplier * config.precisionMultiplier;
+        uint256 price = (config.baseLength * config.maxPrice / length)
```
Whytecrowe: why do we use an extra variable here? this seem to only add unnecessary gas

contracts/utils/StringUtils.sol
```diff
@@ -33,4 +33,28 @@ library StringUtils {
         }
         return len;
     }
+
+    /**
+     * @dev Confirm that a given string has only alphanumeric characters [a-z0-9]
```
Whytecrowe: ... or a hyphen

contracts/utils/StringUtils.sol
```diff
@@ -33,4 +33,28 @@ library StringUtils {
...
+     * @param s The string to validate
+     */
+    function validate(string memory s) internal pure {
+        bytes memory nameBytes = bytes(s);
+        uint256 length = nameBytes.length;
+
+        uint256 MAX_INT = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
```
Whytecrowe, line 45: is this actually max int or max Uint?

contracts/utils/StringUtils.sol
```diff
@@ -33,4 +33,28 @@ library StringUtils {
...
+        require(length < MAX_INT, "ZNSRootRegistrar: Domain name too long");
```
Whytecrowe: wrong contract name in the msg here

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -87,13 +90,18 @@ contract ZNSRootRegistrar is
         string calldata tokenURI,
         DistributionConfig calldata distributionConfig
     ) external override returns (bytes32) {
+        // Confirms string values are only [a-z0-9]
+        name.validate();
```
Whytecrowe, line 94: why do we do it in this function as well? we will do this twice for every register

contracts/price/ZNSCurvePricer.sol
```diff
@@ -61,6 +61,9 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         bytes32 parentHash,
         string calldata label
     ) public view override returns (uint256) {
+        // Confirms string values are only [a-z0-9]
+        label.validate();
```
Whytecrowe: why do we do it in this function as well? we will do this twice for every register

PR #70, opened 2023-11-09
fix: Create the ability to completely configure the environment

JamesEarle:
Add code for custom environments where a default configuration is built but any environment variables that are given have the ability to override the default.

Fixed an issue where hardhat would infer the signer of a `deployProxy` call to always be the first account when it generates on hardhat, and because no accounts were given to the implementation factory specifically it would always use this even if the configuration said otherwise, causing issues with access control.

For [MUD-318](https://wilderworld.atlassian.net/browse/MUD-318?atlOrigin=eyJpIjoiNDhmNWMzNWUxMjY5NDFiYzhkN2IyY2E0MmY5NWNlYzgiLCJwIjoiaiJ9)

[MUD-318]: https://wilderworld.atlassian.net/browse/MUD-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env.sample
```diff
@@ -0,0 +1,38 @@
...
+REG_FEE_PERCENT=
+ROYALTY_FRACTION=
+
+TOKEN_NAME=
+TOKEN_SYMBOL=
+ROYALTY_RECEIVER=
+ZERO_VAULT_ADDRESS=
+STAKING_TOKEN_ADDRESS=
+
+# comma separated list of address e.g. "0x123,0x456,..."
```
Whytecrowe: plz specify that they should be added as base64 string

src/deploy/constants.ts
```diff
@@ -1,6 +1,8 @@
 import { ethers } from "ethers";
 import { IProxyKinds } from "./missions/types";
 
+// Mainnet MEOW token address
+export const MEOW_TOKEN = "0x0ec78ed49c2d27b315d462d43b5bab94d2c79bf8";
```
Whytecrowe: This can be remove in favor of the existing object here:
https://github.com/zer0-os/zNS/blob/feat/deploy-campaign/src/deploy/missions/contracts/meow-token/mainnet-data.ts#L3

src/deploy/db/mongo-adapter/constants.ts
```diff
@@ -11,5 +11,6 @@ export const VERSION_TYPES = {
 };
 
 // TODO dep: move these to default ENV config
-export const mongoURILocal = "mongodb://localhost:27018";
-export const mongoDbName = "zns-campaign";
+export const DEFAULT_MONGO_URI = "mongodb://localhost:27018";
+export const DEFAULT_MONGO_DB_NAME = "zns-campaign";
+export const DEFAULT_MONGO_VERSION = "7.0";
```
Whytecrowe: remove this. in test if the version is not specified, it will create it here:
https://github.com/zer0-os/zNS/blob/feat/deploy-campaign/src/deploy/db/mongo-adapter/mongo-adapter.ts#L129-L132

src/deploy/db/mongo-adapter/get-adapter.ts
```diff
@@ -1,34 +1,38 @@
...
     logger,
-    clientOpts: !!process.env.MONGO_DB_CLIENT_OPTS
+    clientOpts: process.env.MONGO_DB_CLIENT_OPTS
       ? JSON.parse(process.env.MONGO_DB_CLIENT_OPTS)
       : undefined,
-    // TODO dep: add better way to set version ENV var is not the best !
-    version: process.env.MONGO_DB_VERSION,
+    version: process.env.MONGO_DB_VERSION
+      ? process.env.MONGO_DB_VERSION
+      : DEFAULT_MONGO_VERSION,
```
Whytecrowe: L26 and L27 can be removed. see - https://github.com/zer0-os/zNS/pull/70/files#r1394911292

src/deploy/db/mongo-adapter/get-adapter.ts
```diff
@@ -1,34 +1,38 @@
...
   };
 
   if (!checkParams.dbUri && !checkParams.dbName) {
```
Whytecrowe: this will never be hit since you have this - https://github.com/zer0-os/zNS/pull/70/files#diff-eb1e69c1e785f8b52956c5da9f661f19f2fc3f309c61c7373dcf448776ec0410R10-R15

If you keep it this way this whole checking flow can be removed.

src/deploy/deployer/hardhat-deployer.ts
```diff
@@ -1,12 +1,15 @@
...
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 
 export class HardhatDeployer {
   hre : HardhatRuntimeEnvironment;
+  signer : SignerWithAddress;
 
-  constructor () {
+  constructor (signer : SignerWithAddress) {
     this.hre = hre;
+    this.signer = signer;
```
Whytecrowe, line 12: good catch!

.env.sample
JamesEarle: Agreed, this change is already on my other branch [deploy/validate-env](https://github.com/zer0-os/zNS/pull/73/files#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049c), after merging my [`deploy/script`](https://github.com/zer0-os/zNS/pull/74) that PR can be merged into this branch, which resolves the issue

src/deploy/db/mongo-adapter/get-adapter.ts
JamesEarle: Will remove this check

PR #73, opened 2023-11-14
fix: Add environment validation

JamesEarle:
Add verification of several logical rules based on the environment specified and variables given.

For [MUD-337](https://wilderworld.atlassian.net/browse/MUD-337?atlOrigin=eyJpIjoiZDY4YjI2MWE1M2VlNDk3NmIyY2U1Mzk5Y2FkYjU0MWMiLCJwIjoiaiJ9)

[MUD-337]: https://wilderworld.atlassian.net/browse/MUD-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/deploy/campaign/environments.ts
```diff
@@ -117,8 +119,56 @@ export const getConfig = (
...
 };
+
+// For testing the behaviour when we manipulate, we have an optional "env" string param
+export const validate = (config : IDeployCampaignConfig, env ?: string, mongoUri ?: string) => {
+  let envLevel;
+
+  if (env) {
+    envLevel = env;
+  } else {
+    envLevel = process.env.ENV_LEVEL;
```
Whytecrowe: this assumes env var will always be set, but this can be undefined as well. we should check that

src/deploy/campaign/environments.ts
```diff
@@ -117,8 +119,56 @@ export const getConfig = (
...
+  }
+
+  if (envLevel === "dev") return; // No validation needed for dev
+
+  if (!mongoUri) mongoUri = process.env.MONGO_URI ? process.env.MONGO_URI : DEFAULT_MONGO_URI;
+
+  // Mainnet or testnet
+  if (envLevel === "prod" || envLevel === "test") {
+    requires(!config.mockMeowToken, "Cannot mock MEOW token in production");
+    requires(config.stakingTokenAddress === MEOW_TOKEN, "Must use MEOW token in production");
```
Whytecrowe: use this here - https://github.com/zer0-os/zNS/blob/feat/deploy-campaign/src/deploy/missions/contracts/meow-token/mainnet-data.ts#L3

test/ZNSRootRegistrar.test.ts
```diff
@@ -154,6 +161,123 @@ describe("ZNSRootRegistrar", () => {
     );
   });
 
+  it("Throws if env variable is invalid", async () => {
```
Whytecrowe: these should be moved to DeployCampaign.test.ts
this is a test for RootRegistrar contract

src/deploy/campaign/environments.ts
```diff
@@ -117,8 +119,56 @@ export const getConfig = (
...
+    requires(validatePrice(config.rootPriceConfig), "Must use a valid price configuration");
+    requires(!mongoUri.includes("localhost"), "Cannot use local mongo URI in production");
+  }
+
+  // If we reach this code, there is an env variable but it's not valid.
+  throw new Error("Invalid environment value. Must set env to one of `dev`, `test`, or `prod`");
```
Whytecrowe: use the constants you created for this and the above errors. the same constants you use in tests

src/deploy/campaign/environments.ts
```diff
@@ -117,8 +119,56 @@ export const getConfig = (
...
+    stakingTokenAddress: process.env.STAKING_TOKEN_ADDRESS ? process.env.STAKING_TOKEN_ADDRESS : MEOW_TOKEN,
   };
 
+  // Will throw an error based on any invalid setup, given the `ENV_LEVEL` set
+  validate(config);
+
   return config;
 };
+
+// For testing the behaviour when we manipulate, we have an optional "env" string param
```
Whytecrowe: this should be read from the ENV directly. the test should also reset this value in the ENV rather than passing here. the way it is now we are not really testing the environment setup, just that we passed the right argument to this function.

test/ZNSRootRegistrar.test.ts
```diff
@@ -154,6 +161,123 @@ describe("ZNSRootRegistrar", () => {
...
+    try {
+      const config = await getConfig(
+        deployer,
+        zeroVault,
+        [deployer.address, governor.address],
+        [deployer.address, admin.address],
+      );
+
+      validate(config, "other");
```
Whytecrowe: see this: https://github.com/zer0-os/zNS/pull/73/files#r1395017387

Please, set the ENV vars in the proper way here so that the validation test actually tests a real-life scenario. right now a discrepancy is possible where ENV_LEVEL can be set to a different value compared to this argument hardcoded in the test.

test/ZNSRootRegistrar.test.ts
```diff
@@ -154,6 +161,123 @@ describe("ZNSRootRegistrar", () => {
...
+    try {
+      const config = await getConfig(
+        deployer,
+        zeroVault,
+        [deployer.address, governor.address],
+        [deployer.address, admin.address],
+      );
+      // Modify the config
+      config.mockMeowToken = true;
+      validate(config, "prod");
```
Whytecrowe: do the above for all of these.

test/ZNSRootRegistrar.test.ts
```diff
@@ -154,6 +161,123 @@ describe("ZNSRootRegistrar", () => {
...
+        [deployer.address, governor.address],
+        [deployer.address, admin.address],
+      );
+
+      config.mockMeowToken = false;
+      config.stakingTokenAddress = "0x123";
+
+      validate(config, "prod");
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    } catch(e : any) {
```
Whytecrowe: `any` here can be `Error` type, no?

test/ZNSRootRegistrar.test.ts
```diff
@@ -154,6 +161,123 @@ describe("ZNSRootRegistrar", () => {
...
+        deployer,
+        zeroVault,
+        [deployer.address, governor.address],
+        [deployer.address, admin.address],
+      );
+
+      config.mockMeowToken = false;
+
+      // Normally we would call to an env variable to grab this value
+      const uri = "";
```
Whytecrowe: this should check the actual ENV variable as well.

src/deploy/campaign/environments.ts
JamesEarle: The below lines explicitly check for `dev`, `test`, and `prod`, but if none of those are satisfied (the case that you've specified) we call the revert message

I'll add a comment for clarity

src/deploy/campaign/environments.ts
JamesEarle: This is going back to the issue we discussed last week, manipulating `process.env.<VAR_NAME>` isn't always reliable because of the async nature behind the scenes in hardhat and mocha, so we can set it in one test but it will affect the outcome of another test that's entirely unrelated

I thought I'd try it again to confirm behavior and I wanted to also add that it doesn't even respect the manipulation most of the time because what it loads from your actual `.env` file will take precedence over anything you set in code.

I'd argue that manipulating the `envLevel` variable that comes from either of the local test `env` variable, or the `process.env.<VAR_NAME>` is equivalent here, either results in failures here if not set properly and the actual function we care about, `getConfig` does not use these test parameters.

test/ZNSRootRegistrar.test.ts
JamesEarle: Causes a TypeScript error for me `Catch clause variable type annotation must be 'any' or 'unknown' if specified.` so I'd have to override that or the `explicit-any` either way

PR #76, opened 2023-11-17
fix: Add contracts git version to DB

Whytecrowe:


.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           command: yarn check-coverage #Validate code is adequately covered by unit tests
       - persist_to_workspace:
           root: ~/repo
-          paths: .
+          paths: [.]
```
JamesEarle, line 31: did you confirm that this is the right syntax here? not something like `["."]` instead, for example

src/deploy/db/mongo-adapter/mongo-adapter.ts
```diff
@@ -84,7 +86,9 @@ export class MongoDBAdapter {
 
   // Contract methods
   async getContract (contractName : string, version ?: string) {
-    if (!version) version = await this.getCheckLatestVersion();
+    if (!version) {
+      ({ dbVersion: version } = await this.getCheckLatestVersion());
```
JamesEarle, line 90: nitpick: outer parens dont seem necessary `( ... )`

src/deploy/db/mongo-adapter/mongo-adapter.ts
```diff
@@ -119,10 +125,10 @@ export class MongoDBAdapter {
     if (version) {
       finalVersion = version;
 
-      if (version !== deployedV || !deployedV) {
+      if (version !== deployedV?.dbVersion || !deployedV) {
```
JamesEarle: reordering this if and others like it will make it so you don't have to use the nullable `?` because the existence of the `deployedV` variable is checked first

make it `if (!deployedV || version !== deployedV.dbVersion)`

src/deploy/db/mongo-adapter/mongo-adapter.ts
```diff
@@ -93,7 +97,9 @@ export class MongoDBAdapter {
   }
 
   async writeContract (contractName : string, data : IContractDbData, version ?: string) {
-    if (!version) version = await this.getCheckLatestVersion();
+    if (!version) {
+      ({ dbVersion: version } = await this.getCheckLatestVersion());
```
JamesEarle, line 101: same as above with outer parens

src/deploy/db/mongo-adapter/mongo-adapter.ts
```diff
@@ -119,10 +125,10 @@ export class MongoDBAdapter {
     if (version) {
       finalVersion = version;
 
-      if (version !== deployedV || !deployedV) {
+      if (version !== deployedV?.dbVersion || !deployedV) {
         // we should only have a single TEMP version at any given time
-        if (version !== tempV && tempV) {
-          await this.clearDBForVersion(tempV);
+        if (version !== tempV?.dbVersion && tempV) {
```
JamesEarle: same as above, you can reverse the order of the if statement and remove the use of the `?`

.circleci/config.yml
Whytecrowe: the CI is not showing a type issue there anymore. so I guess it is... :)

src/deploy/db/mongo-adapter/mongo-adapter.ts
Whytecrowe: since it's not const, they are. it shows an error otherwise

src/utils/git-tag/save-tag.sh
JamesEarle, line 1: cool solution!

src/deploy/db/mongo-adapter/mongo-adapter.ts
Whytecrowe: you are right! good catch

src/utils/git-tag/save-tag.sh
Whytecrowe: wanted to do this from the get go, but decided to keep to ts. of course bash is more reliable hahah

JamesEarle (approved):

PR #77, opened 2023-11-22
fix: Single integration test for deploy campaign

JamesEarle:
Single integration test for deploy campaign that runs deployment and then also proceeds to run through several crucial flows as part of verification.

[For 339](https://wilderworld.atlassian.net/browse/MUD-339)


package.json
```diff
@@ -45,7 +45,7 @@
     "@nomicfoundation/hardhat-network-helpers": "^1.0.8",
     "@nomicfoundation/hardhat-toolbox": "2.0.2",
     "@nomiclabs/hardhat-ethers": "2.2.2",
-    "@nomiclabs/hardhat-etherscan": "3.0.0",
+    "@nomiclabs/hardhat-etherscan": "3.1.7",
```
Whytecrowe, line 47: you don't get an error with this? I thought this package is already part of the `hardhat-toolbox`, no?

test/DeployCampaign.integration.test.ts
```diff
@@ -0,0 +1,315 @@
...
+    config.mockMeowToken = hre.network.name === "hardhat";
+    const campaign = await runZnsCampaign({ config, dbVersion: "1.0.0" });
+
+    zns = campaign.state.contracts;
+
+    //  CurvePricer, stake, open
+    distConfig = {
+      pricerContract: zns.curvePricer.address,
+      paymentType: 1,
+      accessType: 1,
```
Whytecrowe: please use existing enums already created for each of these fields:

https://github.com/zer0-os/zNS/blob/feat/verify-monitor-contracts/test/helpers/constants.ts#L20-L37

test/DeployCampaign.integration.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      users,
+      domains,
+      domainAddressMock.address,
+      "https://zns.domains/", // tokenUri
+      distConfig,
+      config.rootPriceConfig,
+      zns
+    );
+
+    const balanceAfterPromises = [
```
Whytecrowe, line 150: it can be the same array as before.

test/helpers/deploy-helpers.ts
```diff
@@ -0,0 +1,134 @@
...
+  domains : Array<string>,
+  domainAddress : string,
+  tokenUri : string,
+  distConfig : IDistributionConfig,
+  priceConfig : ICurvePriceConfig,
+  zns : TZNSContractState,
+) : Promise<void> => {
+  let index = 0;
+
+  // TODO need to return state for outer calling tests, better way to do this?
```
Whytecrowe: what is this for? still needs to be done?

package.json
JamesEarle: I had an error without adding this, and this was the resolution. I couldn't setup a Sepolia network anymore in the HH config without upgrading the package

test/helpers/deploy-helpers.ts
JamesEarle: No I resolved that it was more meant to be explanatory. I couldn't get the state of the contracts in the scope of the code that calls `registerRootDomainBulk` to reflect the changes made within that function if I used a prototype function like `.map()` or `.reduce()`, I'm sure there is a way but it wouldn't work for me and I went with this instead. Removed the comment

test/DeployCampaign.integration.test.ts
JamesEarle: I don't think this is true. I tried this to confirm and it only returns the before values when we don't have a second "after" array even if they are all the same promises.

When using a single promise array it fails the test with this error, and when using two different arrays (containing the same promises) it passes the tests. Must be another JavaScript behind the scenes thing, trying to save time and can see that the promises in the array are already resolved and so even with `await Promise.all()` running a second time, it just returns the resolved values
 
image

PR #91, opened 2023-12-12
[DEPLOY MAIN - MERGE FIRST] fix: Changes to use only hre.upgrades in deployment

JamesEarle:
Even with signer and provider given from Defender, we don't use the Defender SDK client to do the actual deployment. Because the Defender SDK doesn't seem to work at the moment, this solution is functional but by navigating around the problem. By not using the Defender SDK, the contracts don't show in Defender even though they are connected to the Defender relayer we've created


src/deploy/missions/contracts/meow-token/meow-token.ts
```diff
@@ -84,7 +84,7 @@ export class MeowTokenDM extends BaseDeployMission {
 
     // Mint 100,000 MEOW to the deployer
     await meowToken.connect(deployAdmin).mint(
-      deployAdmin.address,
+      await deployAdmin.getAddress(),
```
Whytecrowe, line 91: the opposite

src/deploy/run-campaign.ts
```diff
@@ -1,9 +1,12 @@
 import { getConfig } from "./campaign/environments";
-import { getLogger } from "./logger/create-logger";
 import { runZnsCampaign } from "./zns-campaign";
 import { Defender } from "@openzeppelin/defender-sdk";
 
-const logger = getLogger();
+// import { DefenderRelayProvider, DefenderRelaySigner } from "@openzeppelin/defender-sdk-relay-signer-client/lib/ethers";
+
+// import { IDeployCampaignConfig } from "./campaign/types";
+// import { BaseContract, ContractFactory, Signer } from "ethers";
```
Whytecrowe: what's these?

src/deploy/run-campaign.ts
```diff
@@ -32,12 +35,13 @@ const runCampaign = async () => {
...
   await runZnsCampaign({
     config,
+    provider,
   });
 };
 
 runCampaign().catch(error => {
-  logger.error(error.message);
   process.exitCode = 1;
+  console.log(error);
```
Whytecrowe: revert this back plz

PR #96, opened 2024-01-10
feat: Add a `protocolFee` to the unstake action

JamesEarle:
This creates a more symmetric system as we charge the `protocolFee` on both the staking and unstaking of a domain.

### DIFF
--- a/contracts/treasury/ZNSTreasury.sol
+++ b/contracts/treasury/ZNSTreasury.sol
@@ -156,13 +156,21 @@ contract ZNSTreasury is AAccessControlled, ARegistryWired, UUPSUpgradeable, IZNS
      * Since we are clearing storage, gas refund from this operation makes Revoke transactions cheaper.
      * @param domainHash The hash of the domain for which the stake is being withdrawn.
      * @param owner The address of the user who is withdrawing the stake.
+     * @param protocolFee The protocol fee paid by the user to Zero.
      */
     function unstakeForDomain(
         bytes32 domainHash,
-        address owner
+        address owner,
+        uint256 protocolFee
     ) external override onlyRegistrar {
         Stake memory stakeData = stakedForDomain[domainHash];
         delete stakedForDomain[domainHash];
+        
+        stakeData.token.safeTransferFrom(
+            owner,
+            paymentConfigs[0x0].beneficiary,
+            protocolFee
+        );
 
         stakeData.token.safeTransfer(owner, stakeData.amount);
 
--- a/contracts/treasury/ZNSTreasury.sol
+++ b/contracts/treasury/ZNSTreasury.sol
@@ -165,7 +165,7 @@ contract ZNSTreasury is AAccessControlled, ARegistryWired, UUPSUpgradeable, IZNS
     ) external override onlyRegistrar {
         Stake memory stakeData = stakedForDomain[domainHash];
         delete stakedForDomain[domainHash];
-        
+
         if (protocolFee != 0) {
             stakeData.token.safeTransferFrom(
                 owner,
--- a/contracts/treasury/ZNSTreasury.sol
+++ b/contracts/treasury/ZNSTreasury.sol
@@ -166,7 +166,7 @@ contract ZNSTreasury is AAccessControlled, ARegistryWired, UUPSUpgradeable, IZNS
         Stake memory stakeData = stakedForDomain[domainHash];
         delete stakedForDomain[domainHash];
 
-        if (protocolFee != 0) {
+        if (protocolFee > 0) {
             stakeData.token.safeTransferFrom(
                 owner,
                 paymentConfigs[0x0].beneficiary,
--- a/contracts/treasury/ZNSTreasury.sol
+++ b/contracts/treasury/ZNSTreasury.sol
@@ -166,11 +166,13 @@ contract ZNSTreasury is AAccessControlled, ARegistryWired, UUPSUpgradeable, IZNS
         Stake memory stakeData = stakedForDomain[domainHash];
         delete stakedForDomain[domainHash];
         
-        stakeData.token.safeTransferFrom(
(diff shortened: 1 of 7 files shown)


contracts/treasury/ZNSTreasury.sol
```diff
@@ -156,13 +156,21 @@ contract ZNSTreasury is AAccessControlled, ARegistryWired, UUPSUpgradeable, IZNS
...
     function unstakeForDomain(
         bytes32 domainHash,
-        address owner
+        address owner,
+        uint256 protocolFee
     ) external override onlyRegistrar {
         Stake memory stakeData = stakedForDomain[domainHash];
         delete stakedForDomain[domainHash];
+        
+        stakeData.token.safeTransferFrom(
```
Whytecrowe: what if this fee is 0?
Then we are paying for a transfer of 0 amount.
Probably needs a check.

test/ZNSRootRegistrar.test.ts
```diff
@@ -984,14 +984,48 @@ describe("ZNSRootRegistrar", () => {
...
       // Verify final balances
-      const computedFinalBalance = balance + staked;
+      const computedFinalBalance = balance + staked - protocolFee;
       const finalBalance = await zns.meowToken.balanceOf(user.address);
       expect(computedFinalBalance).to.equal(finalBalance);
     });
   });
 
   describe("Revoking Domains", () => {
+    it("Sends a protocol fee as part of revoke flow", async () => {
```
Whytecrowe: a bit unclear what this means. maybe smth like - "Charges a protocol fee from owner on revoke".

test/ZNSRootRegistrar.test.ts
```diff
@@ -984,14 +984,48 @@ describe("ZNSRootRegistrar", () => {
...
+      const domainHash = await getDomainHashFromEvent({
+        zns,
+        user,
+      });
+
+      const [ price, stakeFee ] = await zns.curvePricer.getPriceAndFee(ethers.ZeroHash, defaultDomain, false);
+
+      const balanceBefore = await zns.meowToken.balanceOf(user.address);
+
+      // is revoke meant to be free if owner of parent? register subdomain is
```
Whytecrowe, line 1018: free in what sense? and why is this a question here?

test/ZNSRootRegistrar.test.ts
```diff
@@ -984,14 +984,48 @@ describe("ZNSRootRegistrar", () => {
...
+          accessType: AccessType.OPEN,
+        },
+      });
+
+      const domainHash = await getDomainHashFromEvent({
+        zns,
+        user,
+      });
+
+      const [ price, stakeFee ] = await zns.curvePricer.getPriceAndFee(ethers.ZeroHash, defaultDomain, false);
```
Whytecrowe: `protocolFee`, not `stakeFee`. we shouldn't confuse ourselves since we use these terms for different things.

`stakeFee` is set by the parent for the subdomains. here you are working with root domain, so there's no `stakeFee` possible.

test/ZNSRootRegistrar.test.ts
```diff
@@ -984,14 +984,48 @@ describe("ZNSRootRegistrar", () => {
...
+      const balanceBefore = await zns.meowToken.balanceOf(user.address);
+
+      // is revoke meant to be free if owner of parent? register subdomain is
+      await zns.rootRegistrar.connect(user).revokeDomain(domainHash);
+
+      const balanceAfter = await zns.meowToken.balanceOf(user.address);
+
+      // Stake fee and protocol fee are calculated the same way, so we can use
+      // stake fee here to represent that the difference in balance after revoking
+      // shows a protocol fee was sent as well
```
Whytecrowe: here you already confused yourself. :) we are using the correct `protocolFee` here, since we've acquired if from zero hash. the thing that's wrong here is the naming.

test/ZNSSubRegistrar.test.ts
```diff
@@ -1516,9 +1536,25 @@ describe("ZNSSubRegistrar", () => {
...
 
       expect(contractBalAfter - contractBalAfterRevoke).to.eq(expectedPrice);
-      expect(childBalAfterRevoke - childBalAfter).to.eq(expectedPrice);
+      expect(childBalAfterRevoke - childBalAfter).to.eq(expectedPrice - protocolFeeOut);
       expect(parentBalAfterRevoke - parentBalAfter).to.eq(0);
-      expect(zeroVaultBalanceAfterRevoke - zeroVaultBalanceAfter).to.eq(0);
+      expect(zeroVaultBalanceAfterRevoke - zeroVaultBalanceAfter - protocolFeeOut).to.eq(0);
+    });
+
+    it("Does not charge the owner of a parent domain when they revoke a sbudomain", async () => {
```
Whytecrowe: typo

Whytecrowe (changes requested):
please, address comments

contracts/treasury/ZNSTreasury.sol
JamesEarle: We only call `unstakeForDomain` if in the `_coreRevoke` flow the `stakedAmount > 0` and so this will always be non-zero, unless we change the fee percentage charged in the root pricer.

test/ZNSRootRegistrar.test.ts
JamesEarle: The value is correct still, it was just an effort to use one `await` call instead of two, to get the same value under a different name.

test/ZNSRootRegistrar.test.ts
JamesEarle: I feel like the comment is explanatory that this is the case and so it's not confusing in my opinion, but I removed this and made it more explicit using two await calls to get the fee properly

Whytecrowe (approved):

PR #99, opened 2024-01-20
[??? POSSIBLY SKIP] Restrict registration for owners of parent domains

JamesEarle:
This makes it so you can only mint free subdomains if you own both the ERC721 token and the domain name in the record held by the ZNSRegistry, solving a potential issue that comes from a token being transferred in a sale on a third party exchange like OpenSea and that new owner not calling to `reclaim` their ownership of the domain. In this case, the previous owner would still be allowed to register subdomains for free.

> SHOULD we edit this to instead use owner of the Token, since we are offering split ownership transfers and regular transfers now switch both owners???
> ANALYZE!


contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -37,6 +37,12 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
         uint256 ownerIndex;
     }
 
+    struct Ownership {
```
Whytecrowe: move to the interface where all the structs are?

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -99,14 +105,30 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
             !registry.exists(domainHash),
             "ZNSSubRegistrar: Subdomain already exists"
         );
+        require(
```
Whytecrowe: why is this necessary? maybe we should check the owner address specifically instead?

test/ZNSSubRegistrar.test.ts
```diff
@@ -139,6 +139,123 @@ describe("ZNSSubRegistrar", () => {
...
+
+      const balanceAfter = await zns.meowToken.balanceOf(lvl2SubOwner.address);
+
+      expect(balanceAfter).to.not.eq(balanceBefore);
+    });
+
+    // disallows an owner from registering for free if they don't own both
+    // disallows an operator from registering for free if the owner does not own both
+    // owner surpasses locked domains
+    // operator surpasses locked domains, maybe already tested these two
```
Whytecrowe: seem like leftovers

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -99,14 +105,30 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+            registry.exists(parentHash),
+            "ZNSSubRegistrar: Parent domain does not exist"
+        );
```
Whytecrowe: is this really necessary?
In the below code, if the owner is 0x0, it will return false for everything and the address will also not match. this can possibly be avoided?

test/gas/gas-costs.json
```diff
@@ -1,4 +1,4 @@
 {
 	"Root Domain Price": "475352",
-	"Subdomain Price": "469054"
+	"Subdomain Price": "478990"
```
Whytecrowe: are we sure this is correct? only +9k from all those changes?

Whytecrowe (changes requested):
comments

contracts/registrar/ZNSSubRegistrar.sol
JamesEarle: Will do. Arguably `Mintlist` struct should be there too though no?

contracts/registrar/ZNSSubRegistrar.sol
JamesEarle: I was mistaken with where the `invalid token id` error was coming from. Initially, this was a short circuit because if the tx fails much deeper in the tx when trying to mint it would be more costly and the intent was to try to be less costly if a user makes a mistake and the tx reverts, but it isn't needed

contracts/registrar/ZNSSubRegistrar.sol
JamesEarle: re: above comment, removing

test/gas/gas-costs.json
JamesEarle: not sure this is 100% accurate, getting differing numbers

PR #101, opened 2024-02-08
fix: Coupon mintlist system [PROTO]

JamesEarle:
This change creates a mintlist system based on a verified signer we control having signed off on each "coupon" that an account can use in registration of a subdomain. If the user presents a message that was not signed by our dedicated coupon signer (TBD what account this is) then it is not considered a valid claim.

What still needs to happen
 - changes to the deployment campaign so the EIP712Helper contract is set properly
 - Decision on what account we use as a signer

Considerations:
 - Using this system relies on the assumption that we first get the owner of a domain to sign a message stating that they allow our coupon signer to sign messages on their behalf. Once we've collected this signature, we can create the data they've provided.
 - There is some deciding to do around what the unique data is in each coupon. We can use the domain label as an intrinsically unique piece of data but there is question about how that might be considered forceful by an owner, allowing a specific user to only mint a specific domain, and so alternatives could be some unique ID but this requires we keep track of what has been used on chain in some mapping
 - Tooling around how we create this data is still to be determined. They could upload a JSON file of addresses, or fill a web form, etc. and we have to intake that data and transform it into proper coupons structured as below
 
 ```
// typescript
 type Coupon = {
  parentHash : string;
  registrantAddress : string;
  subdomainLabel : string;
}


// solidity
struct Coupon {
  bytes32 parentHash;
  address registrantAddress
  string domainLabel;
}
```

### CODE REVIEW

File: contracts/registrar/EIP712Helper.sol [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+    bytes32 private constant COUPON_TYPEHASH = keccak256(
+        "Coupon(bytes32 parentHash,address registrantAddress,string domainLabel)"
+    );
+
+	constructor(
+        string memory name,
+        string memory version
+    ) EIP712(name, version) {}
+
+    function hashCoupon(Coupon memory coupon) public view override returns (bytes32) {
```
Whytecrowe, line 27: can this be a pure function?

contracts/registrar/EIP712Helper.sol
```diff
@@ -0,0 +1,59 @@
...
+
+
+contract EIP712Helper is EIP712, IEIP712Helper {
+    using ECDSA for bytes32;
+
+	// TODO make this real, not the HH rootOwner
+	// idea around creating signer in `hashCoupon` or similar
+	// then storing that data, and in recreation we have to get the address that signed?
+	// how do we bulk sign?
+    address private constant COUPON_SIGNER = 0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65;
```
Whytecrowe, line 16: yeah, this should be a state var.

PR #106, opened 2024-05-07
[ZNS ZChain] String Resolver

MichaelKorchagin:


contracts/resolver/IZNSStringResolver.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.18;
+
+
+interface IZNSStringResolver {
+    /**
+     * @param newString The new domain owner
```
Whytecrowe: change the order of these.
also, `newString` comment is not correct. it's not the owner, it's the string that owner assigns which a domain will resolve to

contracts/resolver/ZNSStringResolver.sol
```diff
@@ -0,0 +1,109 @@
...
+     */
+    function setString(
+        bytes32 domainHash,
+        string calldata newString
+    ) external override {
+        // only owner or operator of the current domain can set the string
+        require(
+            registry.isOwnerOrOperator(domainHash, msg.sender) ||
+            // TODO: decide, whether the registrar can set a string
+            accessController.isRegistrar(msg.sender),
```
Whytecrowe: please, remove the comment and L63. Registrar will not be able to set a string. only owner or operator can.

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,464 @@
...
+  ZNSTreasury,
+} from "../typechain";
+import { DeployCampaign, MongoDBAdapter } from "@zero-tech/zdc";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { DefenderRelayProvider } from "@openzeppelin/defender-sdk-relay-signer-client/lib/ethers";
+import { getConfig } from "../src/deploy/campaign/environments";
+
+
+describe("ZNSStringResolver", () => {
+  describe("One campaign for everybody", () => {
```
Whytecrowe: rename to "Single State Tests"

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,464 @@
...
+        deployer,
+        zeroVault,
+        user,
+        deployAdmin,
+        admin,
+        operator,
+      ] = await hre.ethers.getSigners();
+
+      const campaignConfig = await getConfig({
+        deployer: deployer as unknown as SignerWithAddress,
```
Whytecrowe: this probably doesn't need cast to `unknown`. can just cast `as SignerWithAddress` right away, without `unknown`

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,464 @@
...
+      userBalance = ethers.parseEther("1000000000000000000");
+      await meowToken.mint(user.address, userBalance);
+      await meowToken.connect(user).approve(await treasury.getAddress(), ethers.MaxUint256);
+    });
+
+    after(async () => {
+      await mongoAdapter.dropDB();
+    });
+
+    it("Should not let initialize the contract", async () => {
```
Whytecrowe: add "twice" to the test name.
"Should not let initialize the contract twice"

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,464 @@
...
+    it("Should correctly attach the string to the domain", async () => {
+
+      await rootRegistrar.connect(user).registerRootDomain(
+        domainName,
+        ethers.ZeroAddress,
+        uri,
+        distrConfigEmpty,
+        paymentConfigEmpty,
+      );
+      await stringResolver.connect(user).setString(domainNameHash, "hippopotamus");
```
Whytecrowe: put "hippopotamus" into a var and then compare to this var in `expect`. don't hardcode stuff

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,464 @@
...
+
+      expect(
+        await stringResolver.resolveDomainString(domainNameHash)
+      ).to.eq(
+        "hippopotamus"
```
Whytecrowe: put the var here

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,464 @@
...
+    });
+
+    it("Should setRegistry() using ADMIN_ROLE and emit an event", async () => {
+      await expect(
+        stringResolver.connect(admin).setRegistry(admin.address)
+      )
+        .to.emit(stringResolver, "RegistrySet")
+        .withArgs(admin.address);
+
+      expect(await stringResolver.registry()).to.equal(admin.address);
```
Whytecrowe, line 142: since you have a single state (a single deployment of ZNS), reset the Registry here to the proper address after all the checks are done, otherwise other tests would fail. registry needs to be correct in order for next tests to work properly

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,465 @@
...
+      deployerBalance = ethers.parseEther("1000000000000000000");
+      await meowToken.mint(deployer.address, deployerBalance);
+      await meowToken.connect(deployer).approve(await treasury.getAddress(), ethers.MaxUint256);
+    });
+
+    afterEach(async () => {
+      await mongoAdapter.dropDB();
+    });
+
+    it("Should not allow non-owner address to setString (similar domain and string)", async () => {
```
Whytecrowe, line 259: what does "(similar domain and string)" mean? is there a reason you made the parent and child domain the same string?

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,465 @@
...
+
+      await registrationWithSetup({
+        zns: campaign.state.contracts,
+        user: operator,
+        domainLabel: curStringDomain,
+        domainContent: ethers.ZeroAddress,
+      });
+
+      await expect(
+        stringResolver.connect(user).setString(hashDomainLabel(curStringDomain), curStringDomain)
```
Whytecrowe, line 271: I don't understand what this does and why.

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,465 @@
...
+        curString
+      );
+
+      expect(
+        await stringResolver.resolveDomainString(hash)
+      ).to.equal(curString);
+    });
+
+    it.skip("Should allow REGISTRAR_ROLE to setString and emit event." +
+    "TODO: decide, whether the registrar can set a string  (different domain and string)", async () => {
```
Whytecrowe: don't add "TODOs" to the test name. always add them as comments.
and no, Registrar will not be able to set a string through registration. so you can delete this test altogether AFTER you deleted the registrar line from the Solidity code that checks registrar role in `setString()`

Whytecrowe (changes requested):
Please, address the comments and fix other failing tests in the repo.
Good start!

test/ZNSStringResolver.test.ts
MichaelKorchagin: Just different test params to cover more scenarios

contracts/resolver/ZNSStringResolver.sol
```diff
@@ -0,0 +1,107 @@
...
+ * @title The specific Resolver for ZNS that maps domain hashes to strings.
+ */
+contract ZNSStringResolver is
+    UUPSUpgradeable,
+    AAccessControlled,
+    ARegistryWired,
+    ERC165,
+    IZNSStringResolver {
+
+    mapping(bytes32 domainHash => string resolvedString) internal domainStrings;
```
JamesEarle: the name `domainStrings` could be a bit misleading in that it implies the resolved strings are the domains themselves, not what they resolve to, maybe `resolvedStrings` similar to the name you already are using in the mapping key/value pair

contracts/resolver/ZNSStringResolver.sol
```diff
@@ -0,0 +1,107 @@
...
+     * @dev Sets the string for a domain name hash.
+     * @param domainHash The identifying hash of a domain's name
+     * @param newString The new string to map the domain to
+     */
+    function setString(
+        bytes32 domainHash,
+        string calldata newString
+    ) external override {
+        // only owner or operator of the current domain can set the string
+        require(
```
JamesEarle: Using custom errors is more gas efficient. I know they aren't used in the other contracts but they have already been deployed, so maybe we can update that at some point but for now use them instead of `require` and do the opposite logic of what you're checking here

```solidity
if (!registry.IsOwnerOrOperator(domainHash, msg.sender)) {
    revert CustomErrorName();
}
```

src/deploy/constants.ts
```diff
@@ -28,4 +28,7 @@ export const EXECUTOR_ROLE = ethers.solidityPackedKeccak256(
 
 export const ResolverTypes = {
```
JamesEarle, line 29: this would be better as an enum, then it would resolve the "TODO" comment you have below as these are just given values `0`, `1`, `2` etc. unless you specifically assign them something.

```solidity
export enum ResolverTypes {
  ADDRESS,
  STRING
}
```

test/ZNSStringResolver.test.ts
```diff
@@ -0,0 +1,433 @@
...
+        );
+
+        const upgradeTx = domainToken.connect(operator).upgradeTo(await newStringResolver.getAddress());
+
+        await expect(upgradeTx).to.be.revertedWith(
+          getAccessRevertMsg(operator.address, GOVERNOR_ROLE)
+        );
+      });
+
+      // TODO: Falls on the role. I think, cannot give a REGISTRAR_ROLE to mock "deployAdmin".
```
JamesEarle, line 419: Make sure you take care of any existing TODO comments. If they are already fixed, remove them

src/deploy/constants.ts
Whytecrowe: that would require a Registry contract upgrade...

contracts/resolver/ZNSStringResolver.sol
MichaelKorchagin: Called them resolvedString and resolvedStrings, which I dont like, to be honest. Tell me, if I'm wrong and have to find another name)

PR #109, opened 2024-06-26
Codebase and dependency update

Whytecrowe:


test/ZNSSubRegistrar.test.ts
```diff
@@ -421,7 +421,7 @@ describe("ZNSSubRegistrar", () => {
     });
 
     // ! this value can change based on the block gas limit !
-    it("should register subdomain with a label length of 100000 chars", async () => {
+    it.skip("should register subdomain with a label length of 100000 chars", async () => {
```
cgewecke: Possible you could skip this test for coverage (if it's run as a separate job) using the technique described at link below, without really affecting the coverage measurement:

https://github.com/sc-forks/solidity-coverage/blob/master/docs/advanced.md#skipping-tests

Whytecrowe: This is perfect! thank you!

PR #114, opened 2024-07-27
[3.] New formula for Curve pricer

MichaelKorchagin:
MERGED TO: [`rc/multi-zns-main`](https://github.com/zer0-os/zNS/pull/121)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ docker*.tgz
 
 # We don't ever use the generated manifests
 .openzeppelin
+/.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,4 @@ docker*.tgz
 
 # We don't ever use the generated manifests
 .openzeppelin
-/.vscode
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,4 @@ docker*.tgz
 
 # We don't ever use the generated manifests
 .openzeppelin
-.vscode
+/.vscode
--- a/contracts/price/ZNSCurvePricer.sol
+++ b/contracts/price/ZNSCurvePricer.sol
@@ -16,6 +16,10 @@ import { ARegistryWired } from "../registry/ARegistryWired.sol";
  * At `maxLength` length of the domain name. Price after `maxLength` is fixed and is determined using the formula where `length` = `maxLength`.
  */
 contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, IZNSCurvePricer {
+
+    event StringLengthLogged(uint256 length);
+
+
     using StringUtils for string;
 
     /**
@@ -137,7 +141,7 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         setPrecisionMultiplier(domainHash, priceConfig.precisionMultiplier);
         priceConfigs[domainHash].baseLength = priceConfig.baseLength;
         priceConfigs[domainHash].maxPrice = priceConfig.maxPrice;
-        priceConfigs[domainHash].bendMultiplier = priceConfig.bendMultiplier;
+        priceConfigs[domainHash].curveMultiplier = priceConfig.curveMultiplier;
         priceConfigs[domainHash].maxLength = priceConfig.maxLength;
         setFeePercentage(domainHash, priceConfig.feePercentage);
         priceConfigs[domainHash].isSet = true;
@@ -147,7 +151,7 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         emit PriceConfigSet(
             domainHash,
             priceConfig.maxPrice,
-            priceConfig.bendMultiplier,
+            priceConfig.curveMultiplier,
             priceConfig.maxLength,
             priceConfig.baseLength,
             priceConfig.precisionMultiplier,
@@ -173,6 +177,15 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         if (maxPrice != 0) _validateConfig(domainHash);
 
         emit MaxPriceSet(domainHash, maxPrice);
+    }
+
+    function setCurveMultiplier(
+        bytes32 domainHash,
+        uint256 curveMultiplier
+    ) external override onlyOwnerOrOperator(domainHash) {
+        priceConfigs[domainHash].curveMultiplier = curveMultiplier;
+
+        emit CurveMultiplierSet(domainHash, curveMultiplier);
     }
 
     /**
@@ -297,11 +310,10 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         if (config.baseLength == 0) return config.maxPrice;
         if (length <= config.baseLength) return config.maxPrice;
 
-        if (length > config.maxLength) return
-        (config.baseLength * config.maxPrice / config.baseLength + config.bendMultiplier
-        * (config.maxLength - config.baseLength)) / config.precisionMultiplier * config.precisionMultiplier;
+        if (length > config.maxLength) length = config.maxLength;
 
-        return (config.baseLength * config.maxPrice / config.baseLength + config.bendMultiplier
+
+        return (config.baseLength * config.maxPrice / config.baseLength + config.curveMultiplier
         * (length - config.baseLength)) / config.precisionMultiplier * config.precisionMultiplier;
     }
 
@@ -311,12 +323,10 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
      * for domains. If this validation fails, the parent function will revert.
      * @dev We are checking here for possible incorrect passed values: `maxLength` or `baselength`.
      */
-    function _validateConfig(bytes32 domainHash, uint256 maxLength, uint256 baseLength) internal view {
+    function _validateConfig(bytes32 domainHash) internal view {
         if (priceConfigs[domainHash].maxLength <= priceConfigs[domainHash].baseLength)
             revert InvalidBaseLengthOrMaxLength(
-                domainHash,
-                maxLength,
-                baseLength
+                domainHash
             );
     }
 
--- a/contracts/price/ZNSCurvePricer.sol
+++ b/contracts/price/ZNSCurvePricer.sol
@@ -7,8 +7,6 @@ import { StringUtils } from "../utils/StringUtils.sol";
 import { AAccessControlled } from "../access/AAccessControlled.sol";
 import { ARegistryWired } from "../registry/ARegistryWired.sol";
 
-import { console } from "hardhat/console.sol";
-
 /**
  * @title Implementation of the Curve Pricing, module that calculates the price of a domain
  * based on its length and the rules set by Zero ADMIN.
@@ -144,7 +142,7 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         bytes32 domainHash,
         CurvePriceConfig calldata priceConfig
     ) public override {
-        setPrecisionMultiplier(domainHash, priceConfig.precisionMultiplier);
+        _setPrecisionMultiplier(domainHash, priceConfig.precisionMultiplier);
         priceConfigs[domainHash].baseLength = priceConfig.baseLength;
         priceConfigs[domainHash].maxPrice = priceConfig.maxPrice;
         priceConfigs[domainHash].curveMultiplier = priceConfig.curveMultiplier;
@@ -181,10 +179,8 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         bytes32 domainHash,
         uint256 maxPrice
     ) external override onlyOwnerOrOperator(domainHash) {
-        validateConfig(domainHash);
-
         priceConfigs[domainHash].maxPrice = maxPrice;
-
+        validateConfig(domainHash);
         emit MaxPriceSet(domainHash, maxPrice);
     }
 
@@ -205,10 +201,8 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
         bytes32 domainHash,
         uint256 curveMultiplier
     ) external override onlyOwnerOrOperator(domainHash) {
-        validateConfig(domainHash);
-        
         priceConfigs[domainHash].curveMultiplier = curveMultiplier;
-
+        validateConfig(domainHash);
         emit CurveMultiplierSet(domainHash, curveMultiplier);
     }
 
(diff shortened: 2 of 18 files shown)


contracts/price/ZNSCurvePricer.sol
```diff
@@ -175,22 +179,13 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-     * @param minPrice The minimum price to set in $ZERO
-     */
-    function setMinPrice(
+    function setCurveMultiplier(
         bytes32 domainHash,
-        uint256 minPrice
+        uint256 curveMultiplier
     ) external override onlyOwnerOrOperator(domainHash) {
-        priceConfigs[domainHash].minPrice = minPrice;
+        priceConfigs[domainHash].curveMultiplier = curveMultiplier;
```
Whytecrowe: do you have any limitations on what this value could be?
Are there any values that would break the formula if set as `curveMultiplier`?
If you have limits for this value, it would be good to add checks here that will revert if incorrect value is set.

contracts/price/ZNSCurvePricer.sol
```diff
@@ -218,12 +213,11 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
     /**
      * @notice Set the maximum length of a domain name to which price formula applies.
-     * All domain names (labels) that are longer than this value will cost the fixed price of `minPrice`,
-     * and the pricing formula will not apply to them.
+     * All domain names (labels) that are longer than this value will cost the lowest price at maxLength.
      * Validates the config with the new length.
      * Fires `MaxLengthSet` event.
      * Only domain owner/operator can call this function.
-     * > `maxLength` can be set to 0 to make all domains cost `minPrice`!
+     * > `maxLength` can be set to 0!
```
Whytecrowe: what does this do now? setting `maxLength` to 0

.gitignore
```diff
@@ -20,3 +20,4 @@ docker*.tgz
 
 # We don't ever use the generated manifests
 .openzeppelin
+/.vscode
```
Whytecrowe, line 23: no forward slash here

contracts/price/ZNSCurvePricer.sol
```diff
@@ -317,26 +335,32 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-            revert InvalidConfigCausingPriceSpikes(
-                domainHash,
-                priceConfigs[domainHash].minPrice,
-                prevToMinPrice
+        if (priceConfigs[domainHash].maxLength < priceConfigs[domainHash].baseLength)
+            revert InvalidBaseLengthOrMaxLength(
+                domainHash
+            );
+
+        if (priceConfigs[domainHash].baseLength == 0 && priceConfigs[domainHash].curveMultiplier == 0)
```
Whytecrowe: this seems like the wrong place to put this check

test/helpers/constants.ts
```diff
@@ -2,8 +2,8 @@ import { ethers } from "hardhat";
 import { ICurvePriceConfig } from "../../src/deploy/missions/types";
 
 export const DEFAULT_RESOLVER_TYPE = "address";
-export const ZNS_DOMAIN_TOKEN_NAME = "ZERO ID";
-export const ZNS_DOMAIN_TOKEN_SYMBOL = "ZID";
+export const ZNS_DOMAIN_TOKEN_NAME = "Zero Name Service";
+export const ZNS_DOMAIN_TOKEN_SYMBOL = "ZNS";
```
Whytecrowe, line 6: revert this back to what it was

contracts/price/ZNSCurvePricer.sol
```diff
@@ -12,10 +12,12 @@ import { ARegistryWired } from "../registry/ARegistryWired.sol";
  * @title Implementation of the Curve Pricing, module that calculates the price of a domain
  * based on its length and the rules set by Zero ADMIN.
  * This module uses an asymptotic curve that starts from `maxPrice` for all domains <= `baseLength`.
- * It then decreases in price, using the calculated price function below, until it reaches `minPrice`
- * at `maxLength` length of the domain name. Price after `maxLength` is fixed and always equal to `minPrice`.
+ * It then decreases in price, using the calculated price function below.
+ * At `maxLength` length of the domain name. Price after `maxLength` is fixed
```
Whytecrowe: what is this sentence about?

contracts/price/ZNSCurvePricer.sol
```diff
@@ -24,6 +26,12 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
      */
     uint256 public constant PERCENTAGE_BASIS = 10000;
 
+    /**
+     * @notice Value used as a basis to multiply the multiplier,
+     * since Solidity does not support fractions.
```
Whytecrowe: wrong sentence

contracts/price/ZNSCurvePricer.sol
```diff
@@ -24,6 +26,12 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+     */
+    uint256 public constant MULTIPLIER_BASIS = 1000;
```
Whytecrowe: unclear name. why "basis"?

contracts/price/ZNSCurvePricer.sol
```diff
@@ -176,21 +183,24 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-     * Fires `MinPriceSet` event.
-     * Only domain owner/operator can call this function.
-     * @param domainHash The domain hash to set the `minPrice` for
-     * @param minPrice The minimum price to set in $ZERO
+     * @notice Sets the multiplier for domains calculations
+     * to allow the hyperbolic price curve to be bent all the way to a straight line.
+     * Validates the config with the new multiplier in case where `baseLength` is 0 too.
+     * Fires `CurveMultiplier` event.
+     * Only domain owner can call this function.
+     * > `curveMultiplier` can be set to 0 to set a maximum price for all domains!
```
Whytecrowe: please add all possible values here and how they affect the formula

contracts/price/ZNSCurvePricer.sol
```diff
@@ -233,6 +242,8 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
     ) external override onlyOwnerOrOperator(domainHash) {
         priceConfigs[domainHash].maxLength = length;
 
+        _validateConfig(domainHash);
```
Whytecrowe: called twice here

contracts/price/ZNSCurvePricer.sol
```diff
@@ -290,17 +300,25 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-     * @dev Before we calculate the price, 4 different cases are possible:
+     * @dev Before we calculate the price, 6 different cases are possible:
      * 1. `maxPrice` is 0, which means all subdomains under this parent are free
      * 2. `baseLength` is 0, which means we are returning `maxPrice` as a specific price for all domains
      * 3. `length` is less than or equal to `baseLength`, which means a domain will cost `maxPrice`
-     * 4. `length` is greater than `maxLength`, which means a domain will cost `minPrice`
+     * 4. `length` is greater than `maxLength`, which means a domain will cost price by fomula at `maxLength`
+     * 5. The numerator can be less than the denominator, which is achieved by setting a huge value
+     * for `curveMultiplier` or by decreasing the `baseLength` and `maxPrice`, which means all domains 
+     * which are longer than `baseLength` will be for free.
```
Whytecrowe: remove "for"

contracts/price/ZNSCurvePricer.sol
```diff
@@ -317,26 +335,32 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+     * @param domainHash The domain hash to validate its config
      */
     function _validateConfig(bytes32 domainHash) internal view {
-        uint256 prevToMinPrice = _getPrice(domainHash, priceConfigs[domainHash].maxLength);
-        if (priceConfigs[domainHash].minPrice > prevToMinPrice)
-            revert InvalidConfigCausingPriceSpikes(
-                domainHash,
-                priceConfigs[domainHash].minPrice,
-                prevToMinPrice
+        if (priceConfigs[domainHash].maxLength < priceConfigs[domainHash].baseLength)
```
Whytecrowe: move this to appropriate places

contracts/price/ZNSCurvePricer.sol
MichaelKorchagin: - `curveMultiplier` = 1.000 - default. Makes a canonical hyperbola (regular).
- It can be "0", which makes all domain prices max.
- If it is less than 1.000, then it pulls the bend towards the straight line.
- If it is bigger than 1.000, then it makes bigger slope on the chart.

I'm adding these comments to contract to be clear

test/helpers/constants.ts
MichaelKorchagin: It is still being "Zero Name Service" and "ZNS" in my IDE for some reason

contracts/price/ZNSCurvePricer.sol
```diff
@@ -134,20 +142,20 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
         priceConfigs[domainHash].baseLength = priceConfig.baseLength;
         priceConfigs[domainHash].maxPrice = priceConfig.maxPrice;
-        priceConfigs[domainHash].minPrice = priceConfig.minPrice;
+        priceConfigs[domainHash].curveMultiplier = priceConfig.curveMultiplier;
         priceConfigs[domainHash].maxLength = priceConfig.maxLength;
         setFeePercentage(domainHash, priceConfig.feePercentage);
         priceConfigs[domainHash].isSet = true;
 
-        _validateConfig(domainHash);
+        validateConfig(domainHash);
```
JamesEarle: small nitpick but because this is an internal function using the name `_validateConfig` would be clearer

contracts/price/ZNSCurvePricer.sol
```diff
@@ -134,20 +142,20 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
         CurvePriceConfig calldata priceConfig
     ) public override {
-        setPrecisionMultiplier(domainHash, priceConfig.precisionMultiplier);
+        _setPrecisionMultiplier(domainHash, priceConfig.precisionMultiplier);
         priceConfigs[domainHash].baseLength = priceConfig.baseLength;
         priceConfigs[domainHash].maxPrice = priceConfig.maxPrice;
-        priceConfigs[domainHash].minPrice = priceConfig.minPrice;
+        priceConfigs[domainHash].curveMultiplier = priceConfig.curveMultiplier;
         priceConfigs[domainHash].maxLength = priceConfig.maxLength;
         setFeePercentage(domainHash, priceConfig.feePercentage);
```
JamesEarle: this function will call `validateConfig` on it's own, then it will be called again a few lines below. Do what you did above with `setPrecisionMultiplier`, break it into two functions where the core is internal, and then the public one calls to validate, and then also to the internal function. Then `setPriceConfig` can call to the internal function only and it will avoid reaching `validateConfig` twice.

contracts/price/ZNSCurvePricer.sol
```diff
@@ -159,83 +165,123 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+        uint256 curveMultiplier
     ) external override onlyOwnerOrOperator(domainHash) {
-        priceConfigs[domainHash].minPrice = minPrice;
+        CurvePriceConfig memory config = priceConfigs[domainHash];
+        _validateCurveMultiplier(domainHash, curveMultiplier, config);
+        emit CurveMultiplierSet(domainHash, curveMultiplier);
+    }
 
-        _validateConfig(domainHash);
+    function _validateCurveMultiplier(
```
Whytecrowe: call this `_validateSetCurveMultiplier()`.
A "validate" function can be `view`, this function is not view cause it actually sets the multiplier. confusing name

contracts/price/ZNSCurvePricer.sol
```diff
@@ -159,83 +165,123 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-        uint256 length
+        uint256 baseLength
     ) external override onlyOwnerOrOperator(domainHash) {
-        priceConfigs[domainHash].baseLength = length;
+        CurvePriceConfig memory config = priceConfigs[domainHash];
+        _validateBaseLength(domainHash, baseLength, config);
+        emit BaseLengthSet(domainHash, baseLength);
+    }
+
+    function _validateBaseLength(
```
Whytecrowe: same for all the other functions. They are setter functions, so their name should reflect that

contracts/price/ZNSCurvePricer.sol
```diff
@@ -247,17 +293,25 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+        emit PrecisionMultiplierSet(domainHash, multiplier);
+    }
 
-        priceConfigs[domainHash].precisionMultiplier = multiplier;
+    function _setPrecisionMultiplier(
+        bytes32 domainHash,
+        uint256 multiplier
+    ) internal {
+        if (multiplier == 0 || multiplier > 10**18
+        ) revert InvalidPrecisionMultiplierPassed(domainHash);
```
Whytecrowe: fix formatting here

contracts/price/ZNSCurvePricer.sol
```diff
@@ -315,29 +387,14 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-        if (priceConfigs[domainHash].minPrice > prevToMinPrice)
-            revert InvalidConfigCausingPriceSpikes(
-                domainHash,
-                priceConfigs[domainHash].minPrice,
-                prevToMinPrice
-            );
+        return ((config.baseLength * config.maxPrice * FACTOR_SCALE) / (
+                    config.baseLength * FACTOR_SCALE + config.curveMultiplier * (length - config.baseLength)
+                ))
+        / config.precisionMultiplier * config.precisionMultiplier;
```
Whytecrowe: make more readable formatting here

contracts/registrar/IZNSSubRegistrar.sol
```diff
@@ -64,16 +67,17 @@ interface IZNSSubRegistrar is IDistributionConfig {
...
-        IZNSPricer pricerContract,
-        PaymentType paymentType,
-        AccessType accessType
-    );
+    )
+        external view returns (
+            IZNSPricer pricerContract,
+            PaymentType paymentType,
+            AccessType accessType
+        );
```
Whytecrowe: this is an unnecessary change, it was fine before.

src/deploy/campaign/environments.ts
```diff
@@ -253,6 +258,7 @@ const requires = (condition : boolean, message : string) => {
 };
 
 // No price spike before `minPrice` kicks in at `maxLength`
+// TODO: do we still need this?
 const validatePrice = (config : ICurvePriceConfig) => {
```
Whytecrowe: no. but we do need other validations. we can't leave them to the contract at this stage.
the deploy code should validate all provided values so that contract does not revert in the middle of a campaign run.

PR #117, opened 2024-09-12
[5.] Replacing the meowToken with zToken

MichaelKorchagin:
MERGED TO: [`rc/multi-zns-main`](https://github.com/zer0-os/zNS/pull/121)


package.json
```diff
@@ -16,6 +16,7 @@
     "lint": "yarn lint-sol & yarn lint-ts --no-error-on-unmatched-pattern",
     "clean": "hardhat clean",
     "build": "yarn run clean && yarn run compile && tsc",
+    "reinstall": "yarn cache clean && rm -rf node_modules && yarn",
```
Whytecrowe: this is not added here. this is a locally run script and has nothing to do with this repo

contracts/token/mocks/ZTokenMock.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.26;
+
+import { ZToken } from "@zero-tech/z-token/contracts/ZToken.sol";
+
+
+contract ZTokenMock is ZToken {
```
JamesEarle, line 7: As a mock it would be helpful to have a public `mint` function, similar to what `MeowTokenMock.sol` had already.

hardhat.config.ts
```diff
@@ -98,29 +98,29 @@ const config : HardhatUserConfig = {
...
-      // ],
-      // // Must have to avoid instead failing as `invalid length for result data` error
-      // throwOnCallFailures: false, // not sure if this even works
-    },
-    devnet: {
-      // Add current URL that you spawned if not using automated spawning
-      url: `${process.env.DEVNET_RPC_URL}`,
-      chainId: 1,
-    },
+    // mainnet: {
```
JamesEarle: Should this be uncommented? Why did you comment out all the networks?

src/deploy/campaign/environments.ts
```diff
@@ -113,8 +127,9 @@ export const getConfig = async ({
       defaultRoyaltyFraction: royaltyFraction,
     },
     rootPriceConfig: priceConfig,
+    zTokenConfig: zConfig,
     zeroVaultAddress: zeroVaultAddressConf as string,
-    mockMeowToken: process.env.MOCK_MEOW_TOKEN === "true",
+    mockZToken: process.env.MOCK_Z_TOKEN === "true",
```
JamesEarle: It would make more sense to set this value in the above on L96 that you added where you already check `process.env.MOCK_Z_TOKEN`

src/deploy/campaign/environments.ts
```diff
@@ -88,6 +92,16 @@ export const getConfig = async ({
     zeroVaultAddressConf = process.env.ZERO_VAULT_ADDRESS;
   }
 
+  let zConfig : IZTokenConfig | undefined;
+  if (process.env.MOCK_Z_TOKEN === "true") {
```
JamesEarle: You don't ever need to test `if (A === true)`, you can just do `if (A)`, or `if (!A)`. A third option would also be `if(!!A)` if you really wanted to get the truth value of something explicitly

src/deploy/campaign/environments.ts
```diff
@@ -146,8 +161,8 @@ export const validateEnv = (
   if (envLevel === "dev") return priceConfig;
 
   if (envLevel === "test" || envLevel === "dev") {
-    if (process.env.MOCK_MEOW_TOKEN === "false" && !process.env.STAKING_TOKEN_ADDRESS) {
-      throw new Error("Must provide a staking token address if not mocking MEOW token in `dev` environment");
+    if (process.env.MOCK_Z_TOKEN === "false" && !process.env.STAKING_TOKEN_ADDRESS) {
```
JamesEarle: Similar here and below on L190, in logical statements it would be sufficient to do `if (!process.env.MOCK_Z_TOKEN)`

src/deploy/campaign/types.ts
```diff
@@ -13,8 +12,10 @@ import {
   ZNSRootRegistrar,
   ZNSSubRegistrar,
   ZNSTreasury,
+  ZToken,
+  ZTokenMock,
 } from "../../../typechain";
-import { MeowToken } from "@zero-tech/ztoken/typechain-js";
+import { bigint } from "hardhat/internal/core/params/argumentTypes";
```
JamesEarle: Remove unused imports

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -86,35 +86,36 @@ IZNSContracts
...
+  // it will choose the governon as `admin` argument
+  // and deployAdmin as `minter` and first passed admin as `mintBeneficiary`.
   async deployArgs () : Promise<TDeployArgs> {
-    return [meowTokenName, meowTokenSymbol];
+    const {
+      initialAdminDelay,
+      initialSupplyBase,
+      inflationRates,
+      finalInflationRate,
+    } = this.config.zTokenConfig as unknown as IZTokenConfig;
```
JamesEarle: We don't need to use `as unknown` here, this object complies with the definition of `IZtokenConfig` fine

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -86,35 +86,36 @@ IZNSContracts
...
-  async postDeploy () {
-    const {
-      meowToken,
-      config: {
-        deployAdmin,
-      },
-    } = this.campaign;
-
-    // Mint 100,000 MEOW to the deployer
-    await meowToken.connect(deployAdmin).mint(
```
JamesEarle, line 113: I mentioned re adding a public `mint` function to the mock above, be sure to put this back in place when you do

src/utils/compare-bytecode.ts
```diff
@@ -1,5 +1,5 @@
 
-// TODO dep: this does not work. find a better way and uncomment the call in meow-token.ts
+// TODO dep: this does not work. find a better way and uncomment the call in z-token.ts
 const replacePattern = /a165627a7a72305820.{64}0029.*/g;
```
JamesEarle, line 2: Since this code is pending more work in a different area it would be nicer to remove this and just make a new branch for this work specifically.

hardhat.config.ts
MichaelKorchagin: Someone was on the way of mock deploy. Got an advise to comment everything until we deploy prod

test/ZNSCurvePricer.test.ts
```diff
@@ -50,14 +51,22 @@ describe("ZNSCurvePricer", () => {
       randomAcc,
     ] = await hre.ethers.getSigners();
 
+    const priceConfig : ICurvePriceConfig  = DEFAULT_PRICE_CONFIG;
```
JamesEarle: Small nitpick, doesn't make a huge difference but assigning this constant to a variable and then using it only one time is unnecessary.  Just give `DEFAULT_PRICE_CONFIG` itself to the `deployZNS` command below

test/ZNSCurvePricer.test.ts
```diff
@@ -50,14 +51,22 @@ describe("ZNSCurvePricer", () => {
...
       governorAddresses: [deployer.address],
       adminAddresses: [admin.address],
+      priceConfig,
+      zeroVaultAddress,
+      isTenderlyRun,
     });
 
-    await zns.meowToken.connect(user).approve(await zns.treasury.getAddress(), ethers.MaxUint256);
-    await zns.meowToken.mint(user.address, DEFAULT_PRICE_CONFIG.maxPrice);
+    const userBalanceInitial = await zns.zToken.balanceOf(admin.address) / 2n;
```
JamesEarle: Why `/ 2n` ? This calls back to other comments I've given you in the past, please always comment things to explain what its doing or why, especially for seemingly arbitrary numbers

test/ZNSTreasury.test.ts
```diff
@@ -61,16 +61,21 @@ describe("ZNSTreasury", () => {
...
     };
 
     // give REGISTRAR_ROLE to a wallet address to be calling guarded functions
     await zns.accessController.connect(admin).grantRole(REGISTRAR_ROLE, mockRegistrar.address);
 
     // Give funds to user
-    await zns.meowToken.connect(user).approve(await zns.treasury.getAddress(), ethers.MaxUint256);
-    await zns.meowToken.mint(user.address, ethers.parseEther("50000"));
+    const userBalanceInitial = await zns.zToken.balanceOf(admin.address) / 3n;
+    await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), userBalanceInitial);
```
JamesEarle: This code is strange to me. Why is the admin calling to approve the treasury here? We only need the `user` to approve the treasury for when they call to register below. To simply transfer token from `admin` to `user` we don't need any approval. There should be just two calls, one for the user to approve the treasury, and one for the admin to transfer funds to the user

test/ZNSSubRegistrar.test.ts
```diff
@@ -88,14 +88,21 @@ describe("ZNSSubRegistrar", () => {
         zeroVaultAddress: zeroVault.address,
       });
       // Give funds to users
+      const users = [
+        rootOwner,
+        lvl2SubOwner,
+      ];
+      const totalAdminBalance = await zns.zToken.balanceOf(admin.address);
+      const userBalanceInitial = totalAdminBalance / BigInt(users.length);
+      await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), totalAdminBalance);
```
JamesEarle: Only the person calling to register a domain needs to approve the treasury, like the registrant below is `rootOwner` and they are calling to approve on L105

test/helpers/deploy-helpers.ts
```diff
@@ -30,16 +30,15 @@ export const approveBulk = async (
...
   amount : bigint,
   zns : IZNSContractsLocal | IZNSContracts,
 ) => {
-  for (const signer of signers) {
-    await zns.meowToken.connect(signer).mint(
-      signer.address,
-      amount
-    );
+  for (const user of users) {
+    await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), ethers.MaxUint256);
```
JamesEarle: There is no need for this here. Think about what this call is doing, we're connecting as the `admin` and calling to approve the treasury. We wouldn't need to do this in a loop even if it was correct, but it also isn't needed here at all because we don't require approval just to do the below transfer from `admin` to `user`

test/helpers/deploy/deploy-zns.ts
```diff
@@ -164,25 +168,27 @@ export const deployDomainToken = async (
...
+    zTokenName,
+    zTokenSymbol,
+    governorAddresses[0],
+    INITIAL_ADMIN_DELAY_DEFAULT,
+    deployer.address,
+    adminAddresses[0],
+    INITIAL_SUPPLY_DEFAULT,
+    INFLATION_RATES_DEFAULT,
+    FINAL_INFLATION_RATE_DEFAULT
+  ) as unknown as ZTokenMock;
```
JamesEarle: We don't need `unknown` here. Look at the output from the above `factory,deploy()` and you'll see that it returns an object of type `ZTokenMock` already.

The `as unkown as MyType` syntax is basically a way to tell the compiler to not complain when you're intentionally breaking type safety. If you aren't doing that, you don't need it.

test/helpers/deploy/deploy-zns.ts
```diff
@@ -193,19 +199,16 @@ export const deployMeowToken = async (
...
+    console.log(`${zTokenMockName} deployed at:
                 proxy: ${proxyAddress}
                 implementation: ${impl}`);
   }
 
-  // Mint 10,000 ZERO for self
-  await meowToken.mint(proxyAddress, ethers.parseEther("10000"));
-
-  return meowToken as unknown as MeowTokenMock;
+  return zToken as unknown as ZTokenMock;
```
JamesEarle, line 212: Same here, no need for `as unknown`

src/deploy/campaign/environments.ts
```diff
@@ -88,6 +92,14 @@ export const getConfig = async ({
     zeroVaultAddressConf = process.env.ZERO_VAULT_ADDRESS;
   }
 
+  process.env.MOCK_Z_TOKEN === "true";
```
JamesEarle: My comment before, which I see you used below as `!!process.env.MOCK_Z_TOKEN`, was that because you read this value twice, it would be be better to just store it in a variable to avoid that. 

Also, you only changed the line I left the comment on. Here we still do `if (A === true)` which is not needed.

Only a small nitpick, but good code is comprised of many small habits :)

JamesEarle (changes requested):
Just the last few comments, otherwise looks good 👍

hardhat.config.ts
JamesEarle: Okay. You don't need to change it, but unless you provide `--network [value]` as a parameter to the tests in the CLI (or to any hardhat command) it won't use these details.

Whytecrowe: just comment out PRIVATE_KEY type vars in there so you don't get errors.

src/deploy/campaign/environments.ts
```diff
@@ -113,8 +125,9 @@ export const getConfig = async ({
       defaultRoyaltyFraction: royaltyFraction,
     },
     rootPriceConfig: priceConfig,
+    zTokenConfig: zConfig,
     zeroVaultAddress: zeroVaultAddressConf as string,
-    mockMeowToken: process.env.MOCK_MEOW_TOKEN === "true",
+    mockZToken: !!process.env.MOCK_Z_TOKEN,
```
Whytecrowe: this is not correct. this variable comes in as a string, NOT as a boolean. so adding "!!" here will make it `true` if it's any string other than empty string (e.g. it will return `true` for string "absandabkshjdg" or "false").
instead you should write it as `process.env.MOCK_Z_TOKEN === "true"`. here you are comparing the string you read from the env to the exact string you need.
remeber that string and boolean are very different types!

src/deploy/campaign/environments.ts
```diff
@@ -146,8 +159,8 @@ export const validateEnv = (
   if (envLevel === "dev") return priceConfig;
 
   if (envLevel === "test" || envLevel === "dev") {
-    if (process.env.MOCK_MEOW_TOKEN === "false" && !process.env.STAKING_TOKEN_ADDRESS) {
-      throw new Error("Must provide a staking token address if not mocking MEOW token in `dev` environment");
+    if (!process.env.MOCK_Z_TOKEN && !process.env.STAKING_TOKEN_ADDRESS) {
```
Whytecrowe: this is incorrect as well. at was correct before, why did you change it? see my previous comment about this variable

src/deploy/campaign/environments.ts
```diff
@@ -172,16 +185,16 @@ export const validateEnv = (
 
   // Mainnet
   if (envLevel === "prod") {
-    requires(process.env.MOCK_MEOW_TOKEN === "false", NO_MOCK_PROD_ERR);
-    requires(process.env.STAKING_TOKEN_ADDRESS === MeowMainnet.address, STAKING_TOKEN_ERR);
     requires(!process.env.MONGO_DB_URI.includes("localhost"), MONGO_URI_ERR);
+    requires(!process.env.MOCK_Z_TOKEN, NO_MOCK_PROD_ERR);
```
Whytecrowe: same error again

Whytecrowe: if you do want to check it this way (with "!") you need to turn env var string into a boolean by `=== "true"` comparison first, then save it to a variable as BOOLEAN, then you can use this variable with "!" everywhere you are now

src/deploy/campaign/environments.ts
```diff
@@ -172,16 +185,16 @@ export const validateEnv = (
...
+    requires(process.env.STAKING_TOKEN_ADDRESS === MeowMainnet.address, STAKING_TOKEN_ERR);
```
Whytecrowe: MeowMainnet.address does not matter anymore. you are using a different token here now

src/deploy/campaign/environments.ts
```diff
@@ -172,16 +185,16 @@ export const validateEnv = (
...
   }
 
-  if (process.env.VERIFY_CONTRACTS === "true") {
+  if (process.env.VERIFY_CONTRACTS) {
```
Whytecrowe: again. this is NOT a boolean. it is a STRING, and you are checking if the string exists, not if it's equal to "true"

src/deploy/campaign/environments.ts
```diff
@@ -172,16 +185,16 @@ export const validateEnv = (
...
     requires(!!process.env.ETHERSCAN_API_KEY, "Must provide an Etherscan API Key to verify contracts");
   }
 
-  if (process.env.MONITOR_CONTRACTS === "true") {
+  if (process.env.MONITOR_CONTRACTS) {
```
Whytecrowe: same thing again

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -4,32 +4,32 @@ import {
...
-import { MeowToken__factory } from "@zero-tech/ztoken/typechain-js";
-import meowArtifact from "@zero-tech/ztoken/artifacts/contracts/MeowToken.sol/MeowToken.json";
+import { ZToken__factory } from "../../../../../typechain";
+import zArtifact from "../../../../../artifacts/@zero-tech/z-token/contracts/ZToken.sol/ZToken.json";
+import { IZTokenConfig } from "../../types";
 
 
-export const meowTokenName = "MEOW";
-export const meowTokenSymbol = "MEOW";
+export const zTokenName = "Z-Token";
```
Whytecrowe: "ZERO Token"

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -4,32 +4,32 @@ import {
...
+export class ZTokenDM extends BaseDeployMission<
 HardhatRuntimeEnvironment,
 SignerWithAddress,
 IZNSCampaignConfig<SignerWithAddress>,
 IZNSContracts
 > {
   proxyData = {
-    isProxy: true,
+    isProxy: false,
     kind: ProxyKinds.transparent,
```
Whytecrowe: remove this line (kind). if it's not a proxy, this is confusing and not needed.

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -39,16 +39,16 @@ IZNSContracts
   >) {
     super(args);
 
-    if (this.config.mockMeowToken) {
-      this.contractName = znsNames.meowToken.contractMock;
+    if (this.config.mockZToken) {
```
Whytecrowe, line 36: now here it is a boolean already, so you can check it like this without problems. no change is needed here, just wanted to point out an example

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -4,32 +4,32 @@ import {
...
 import { ProxyKinds } from "../../../constants";
-import { ethers } from "ethers";
 import { znsNames } from "../names";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { SignerWithAddress } from "@nomicfoundation/hardhat-ethers/signers";
 import { IZNSCampaignConfig, IZNSContracts } from "../../../campaign/types";
-import { MeowToken__factory } from "@zero-tech/ztoken/typechain-js";
-import meowArtifact from "@zero-tech/ztoken/artifacts/contracts/MeowToken.sol/MeowToken.json";
+import { ZToken__factory } from "../../../../../typechain";
+import zArtifact from "../../../../../artifacts/@zero-tech/z-token/contracts/ZToken.sol/ZToken.json";
```
Whytecrowe: this should be removed. we did it before because the other repo was bugged.

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -86,35 +86,36 @@ IZNSContracts
   }
 
   getArtifact () {
-    return meowArtifact;
+    return zArtifact;
```
Whytecrowe: remove this. we did it before because the other repo was bugged.
This repo should work fine, so you can remove this method completely. HardhatDeployer should be able to get the artifact where needed in it's default way.

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -86,35 +86,36 @@ IZNSContracts
...
   }
 
+  // it will choose the governon as `admin` argument
+  // and deployAdmin as `minter` and first passed admin as `mintBeneficiary`.
```
Whytecrowe: no. all these addresses should be parts of the config. we need to be able to specify these addresses as needed.

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -86,35 +86,36 @@ IZNSContracts
...
+    const {
+      initialAdminDelay,
+      initialSupplyBase,
+      inflationRates,
+      finalInflationRate,
+    } = this.config.zTokenConfig as IZTokenConfig;
+    return [
+      zTokenName,
+      zTokenSymbol,
+      this.config.governorAddresses[0],
```
Whytecrowe: destructure this properly as you did with other values. also, these specific role based addresses should be part of the config for zToken

src/deploy/missions/contracts/z-token/z-token.ts
```diff
@@ -86,35 +86,36 @@ IZNSContracts
...
+      this.config.adminAddresses[0],
+      initialSupplyBase,
+      inflationRates,
+      finalInflationRate,
+    ];
   }
 
   async needsPostDeploy () {
-    const msg = this.config.mockMeowToken ? "needs" : "doesn't need";
+    const msg = this.config.ZToken ? "needs" : "doesn't need";
```
Whytecrowe: error here. what is `ZToken` in your config?
this whole method can be safely removed

src/deploy/missions/types.ts
```diff
@@ -8,3 +8,10 @@ export interface ICurvePriceConfig {
   feePercentage : bigint;
   isSet : boolean;
 }
+
+export interface IZTokenConfig {
+  initialAdminDelay : bigint;
+  initialSupplyBase : bigint;
+  inflationRates : Array<bigint>;
+  finalInflationRate : bigint;
```
Whytecrowe, line 21: add all the constructor arguments for the contract here

test/DeployCampaignInt.test.ts
```diff
@@ -69,100 +73,84 @@ describe("Deploy Campaign Test", () => {
...
+          zeroVaultAddress: zeroVault.address,
+          env,
+        }
+      );
+      campaignConfig.domainToken.defaultRoyaltyReceiver = deployAdmin.address;
+      campaignConfig.postDeploy.tenderlyProjectSlug = "";
     });
 
-    it("should deploy new MeowTokenMock when `mockMeowToken` is true", async () => {
+    it("should deploy new ZTokenMock when `mockZToken` is true", async () => {
```
Whytecrowe, line 92: this test is not testing anything. it needs to test what you have in the name.
there is no check for the variable value or that the deployed contract is in fact a mock

test/DeployCampaignInt.test.ts
```diff
@@ -870,27 +849,20 @@ describe("Deploy Campaign Test", () => {
...
+          governors: [deployAdmin.address, governor.address],
+          admins: [deployAdmin.address, admin.address],
+          zeroVaultAddress: zeroVault.address,
+          env,
+        }
+      );
+
+      campaignConfig.domainToken.defaultRoyaltyReceiver = deployAdmin.address;
+      // TODO dep: what do we pass here for test flow? we don't have a deployed ZToken contract
+      campaignConfig.stakingTokenAddress = MeowMainnet.address;
```
Whytecrowe: why are you passing Meow address here?

test/DeployCampaignInt.test.ts
```diff
@@ -1051,27 +1023,19 @@ describe("Deploy Campaign Test", () => {
...
+        {
+          deployer: deployAdmin,
+          governors: [deployAdmin.address, governor.address],
+          admins: [deployAdmin.address, admin.address],
+          zeroVaultAddress: zeroVault.address,
+          env,
+        }
+      );
+      config.domainToken.defaultRoyaltyReceiver = deployAdmin.address;
+      config.stakingTokenAddress = MeowMainnet.address;
```
Whytecrowe: again

test/ZNSCurvePricer.test.ts
```diff
@@ -50,14 +51,22 @@ describe("ZNSCurvePricer", () => {
...
+    const zeroVaultAddress : string = deployer.address;
+    const isTenderlyRun  = false;
+
     zns = await deployZNS({
       deployer,
       governorAddresses: [deployer.address],
       adminAddresses: [admin.address],
+      priceConfig: DEFAULT_PRICE_CONFIG,
+      zeroVaultAddress,
+      isTenderlyRun,
```
Whytecrowe: you don't need to pass this

test/ZNSSubRegistrar.test.ts
Whytecrowe: I agree. does admin here actually needs to approve treasury? is he registering any domains in the test?

test/ZNSSubRegistrar.test.ts
```diff
@@ -2968,18 +2997,25 @@ describe("ZNSSubRegistrar", () => {
...
+        rootOwner,
+        lvl2SubOwner,
+        lvl3SubOwner,
+        lvl4SubOwner,
+        lvl5SubOwner,
+        lvl6SubOwner,
+      ];
+      const totalAdminBalance = await zns.zToken.balanceOf(admin.address);
+      const userBalanceInitial = totalAdminBalance / BigInt(users.length);
+      await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), totalAdminBalance);
```
Whytecrowe: why is this approval here? where is it needed?

test/ZNSSubRegistrar.test.ts
```diff
@@ -3481,14 +3517,21 @@ describe("ZNSSubRegistrar", () => {
       });
 
       // Give funds to users
+      const users = [
+        rootOwner,
+        lvl2SubOwner,
+      ];
+      const totalAdminBalance = await zns.zToken.balanceOf(admin.address);
+      const userBalanceInitial = totalAdminBalance / BigInt(users.length);
+      await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), totalAdminBalance);
```
Whytecrowe: again

test/ZNSTreasury.test.ts
Whytecrowe: agree. there are a lot of admin approvals of registry in this whole file. I believe none of them are needed.

test/ZNSTreasury.test.ts
```diff
@@ -61,16 +61,21 @@ describe("ZNSTreasury", () => {
...
+    await zns.zToken.connect(admin).transfer(user, userBalanceInitial);
+    await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), userBalanceInitial);
```
Whytecrowe: again

test/ZNSTreasury.test.ts
```diff
@@ -628,6 +633,8 @@ describe("ZNSTreasury", () => {
       const newHash = hashSubdomainName(newLabel);
       const { expectedPrice, stakeFee } = getPriceObject(newLabel, DEFAULT_PRICE_CONFIG);
 
+      await zns.zToken.connect(deployer).approve(await zns.treasury.getAddress(), ethers.MaxInt256);
```
Whytecrowe, line 633: again

test/gas/TransactionGasCosts.test.ts
```diff
@@ -51,14 +51,20 @@ describe("Transaction Gas Costs Test", () => {
     };
 
     // Give funds to users
+    const users = [
+      rootOwner,
+      lvl2SubOwner,
+    ];
+    const totalAdminBalance = await zns.zToken.balanceOf(admin.address);
+    const userBalanceInitial = totalAdminBalance / BigInt(users.length);
+    await zns.zToken.connect(admin).approve(await zns.treasury.getAddress(), totalAdminBalance);
```
Whytecrowe: again

test/helpers/deploy-helpers.ts
Whytecrowe: yep

Whytecrowe (changes requested):
some more comments for you

src/deploy/campaign/environments.ts
MichaelKorchagin: > Also, you only changed the line I left the comment on. Here we still do if (A === true) which is not needed.

Didn't understand what do you mean. I replaced this construction ( if (A === true) ) with simple if (A) and don't see any other.

test/ZNSTreasury.test.ts
MichaelKorchagin: We need an approval here, because the deployer is the participant of stakeForDomain function

Whytecrowe (approved):

Whytecrowe:
This was done well, but needs to be closed, since the code has changed too much. Code from this was used in a different PR - https://github.com/zer0-os/zNS/pull/125

PR #118, opened 2024-09-19
[DEFERRED] [Multi ZNS] Multi zNS Prototype [WIP]

Whytecrowe:
`rc/multi-zns-main` needs to be merged into this branch to update the code with all the previous changes !!!


contracts/cross-chain/IZNSZChainPortal.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    function destNetworkId() external view returns (uint32);
+
+    function destChainName() external view returns (string memory);
+
+    function destChainId() external view returns (uint256);
+
+    function destZnsPortal() external view returns (address);
+
+    function rootRegistrar() external view returns (IZNSRootRegistrarTrunk);
```
JamesEarle, line 45: Consider moving this and a few of the other items that are shared between this contract and `IZNSEthereumPortal` to a higher level `IZNSPortal`, to avoid code duplication

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -0,0 +1,221 @@
...
+import { AAccessControlled } from "../access/AAccessControlled.sol";
+import { IZNSZChainPortal } from "./IZNSZChainPortal.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IZNSChainResolver } from "../resolver/IZNSChainResolver.sol";
+import { IZNSPricer } from "../types/IZNSPricer.sol";
+import { IZNSRegistry } from "../registry/IZNSRegistry.sol";
+import { IZNSRootRegistrarTrunk } from "../registrar/IZNSRootRegistrarTrunk.sol";
+import { IZNSSubRegistrarTrunk } from "../registrar/IZNSSubRegistrarTrunk.sol";
+import { IZNSTreasury } from "../treasury/IZNSTreasury.sol";
+import { PaymentConfig } from "../treasury/IZNSTreasury.sol";
```
JamesEarle, line 15: same import file as what's used above, just do `import { X, Y } from .... `

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -0,0 +1,221 @@
...
+            parentHash: parentHash,
+            label: label,
+            tokenUri: tokenURI
+        });
+
+        bytes memory encodedProof = abi.encode(proof);
+
+        // TODO multi: should we leave this? We can't rely on setting this properly on time
+        //  and can permanently lock domains !!!
+        if (destZnsPortal == address(0)) revert DestinationPortalNotSetInState();
```
JamesEarle, line 195: Yes, we should leave this. If there is a possibility that `destZnsPortal` isn't set when we call to bridge, it should revert here

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -0,0 +1,221 @@
...
+
+        polygonZkEVMBridge.bridgeMessage(
+            destNetworkId,
+            destZnsPortal,
+            // TODO multi: figure out what this is and how to better pass it !!!
```
JamesEarle, line 200: is this comment in reference to a similar comment for `forceGlobalExitRoot` above on L74?

contracts/cross-chain/ZNSZChainPortal.sol
JamesEarle, line 1: A bit confusing having the opposite chain in the name of the contract to me. This is the portal to ZChain, but it is not the portal *on* ZChain, and calling it `ZChainPortal` gives the opposite implication

contracts/cross-chain/IZNSZChainPortal.sol
```diff
@@ -0,0 +1,78 @@
...
+import { IZNSSubRegistrarTrunk } from "../registrar/IZNSSubRegistrarTrunk.sol";
+import { IZNSRegistry } from "../registry/IZNSRegistry.sol";
+import { IZNSTreasury } from "../treasury/IZNSTreasury.sol";
+import { IZNSChainResolver } from "../resolver/IZNSChainResolver.sol";
+import { IZNSAccessController } from "../access/IZNSAccessController.sol";
+
+
+interface IZNSZChainPortal is IDistributionConfig {
+
+    struct ZNSContractInput {
```
JamesEarle, line 16: You put this in a struct but the other contract has the same arguments, except for `IZNSAccessController`, might make more sense to use this struct as an initializer argument for both contracts and just have the one additional argument of the access controller contract be given in the `ZNSZChainPortal` contract

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -0,0 +1,221 @@
...
+                parentHash,
+                label,
+                address(0),
+                tokenURI,
+                emptyDistrConfig,
+                emptyPaymentConfig
+            );
+        }
+
+        // TODO multi: analyze how to make these strings better !!!
```
JamesEarle, line 102: Enums?

```solidity
enum ResolverTypes {
  CHAIN,
  ADDRESS,
  STRING
}
```

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -0,0 +1,221 @@
...
+        registry.updateDomainResolver(domainHash, "chain");
+        // TODO multi: iron out what should go to ChainResolver data !!!
+        chainResolver.setChainData(
+            domainHash,
+            destChainId,
+            destChainName,
+            address(0),
```
JamesEarle, line 109: Do we not care about setting `znsRegistryOnChain`?

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -0,0 +1,221 @@
...
+
+        emit DestZnsPortalSet(newAddress);
+    }
+
+    function _processPayment(
+        string calldata label,
+        bytes32 parentHash
+    ) internal {
+        // TODO multi: do we actually want to stake on BOTH sides? What other payment option can we do ???!!!
+        // TODO multi: if we leave this option (stake as Portal address) this needs to be optimized!!!
```
JamesEarle, line 144: Can we not include the registrant address and have the portal use stake payment on the registrant's behalf?

contracts/cross-chain/mock/PolygonZkEVMBridgeV2Mock.sol
```diff
@@ -0,0 +1,161 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.26;
+
+// This is here only to bring the contract so Hardhat compiles it by default
+import { IBridgeMessageReceiver } from "@zero-tech/zkevm-contracts/contracts/interfaces/IBridgeMessageReceiver.sol";
+import { ERC20Mock } from "../../token/mocks/ERC20Mock.sol";
+import { ReentrancyGuardUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/ReentrancyGuardUpgradeable.sol";
+// import { PolygonZkEVMBridgeV2 } from "@zero-tech/zkevm-contracts/contracts/v2/PolygonZkEVMBridgeV2.sol";
```
JamesEarle, line 8: delete unneeded comments

contracts/cross-chain/mock/PolygonZkEVMBridgeV2Mock.sol
```diff
@@ -0,0 +1,161 @@
...
+                    (originAddress, originNetwork, metadata)
+                )
+            );
+        }
+
+        if (!success) {
+            revert MessageFailed();
+        }
+
+        emit ClaimEvent(
```
JamesEarle, line 153: Think of a better name for the events here and above. `MessageClaimed` and `MessageBridged` or just `Claimed` and `Bridged` would be better

src/environment/set-env.ts
```diff
@@ -0,0 +1,12 @@
...
+import { environment } from "./env";
+
+export const setDefaultEnvironment = () => {
+  Object.entries(environment).forEach(([key, value]) => {
+    // load the defaults first
+    process.env[key] = value;
+  });
+
+  // override from local .env file if anything is present
+  require("dotenv").config({ override: true });
```
JamesEarle, line 11: This part might be doing nothing, because env vars are loaded at run time, when a script that uses this function is called, all the `process.env` vars are mapped, then we override those with the defaults above in the `Object.entries` loop

src/environment/validate.ts
```diff
@@ -0,0 +1,20 @@
...
+  "DECIMALS", "PRECISION", "DOMAIN_TOKEN_NAME", "DOMAIN_TOKEN_SYMBOL", "ROYALTY_FRACTION", "MOCK_MEOW_TOKEN",
+  "SRC_CHAIN_NAME", "MOCK_ZKEVM_BRIDGE",
+];
+
+// TODO multi: is there a way to evaluate a type of the full ENV and run the below function
+//  only on non-optional parameters?
+export const findMissingEnvVars = () => {
+  const missing = required.filter(
+    key =>
+      process.env[key] === undefined || process.env[key] === ""
```
JamesEarle, line 14: small nitpick, you often add line breaks at arbitrary points that don't really make sense to me, like here, where it clearly isn't going past the character limit specified by the linter, whereas above in the `required` array or below on L18 you aren't using the same rule

Ideally we just abide by the rules specified in the linter so that we're both always on the same page, and if you prefer to shorten the max line length enforced we can do that instead

hardhat.config.ts
```diff
@@ -121,26 +123,29 @@ const config : HardhatUserConfig = {
...
+  },
+  etherscan: {
+    apiKey: `${process.env.ETHERSCAN_API_KEY}`,
+    customChains: [
+      {
+        network: "zchaintest",
+        chainId: 2012605151,
+        urls: {
+          apiURL: "https://wilderworld-dev-erigon1-blockscout.eu-north-2.gateway.fm/api/",
+          browserURL: "https://wilderworld-dev-erigon1-blockscout.eu-north-2.gateway.fm/",
```
JamesEarle, line 161: put these values in env vars

contracts/registrar/ZNSSubRegistrarTrunk.sol
```diff
@@ -91,57 +92,48 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
         DistributionConfig calldata distrConfig,
         PaymentConfig calldata paymentConfig
     ) external override returns (bytes32) {
+        return _coreSubdomainRegister(
```
JamesEarle, line 95: Here and in other places it would be nicer if the internal `_core` functions were consistent with the naming of the public callers as much as they can be, so `_coreRegisterSubdomain`, this is clearer as well as the name describes the action the function is taking

contracts/registrar/ZNSRootRegistrarBase.sol
```diff
@@ -182,7 +166,7 @@ contract ZNSRootRegistrar is
             registry.createDomainRecord(args.domainHash, args.registrant, "address");
 
             IZNSAddressResolver(registry.getDomainResolver(args.domainHash))
-                .setAddress(args.domainHash, args.domainAddress);
+            .setAddress(args.domainHash, args.domainAddress);
```
JamesEarle, line 169: For formatting like this where lines from a single function call have to broken into multiple for the sake of line length, it makes more sense to have the extra indentation here. It shows visually that this line is not independent, I would undo this change and any other similar changes

contracts/registrar/ZNSSubRegistrarTrunk.sol
```diff
@@ -152,7 +144,48 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+        bool isOwnerOrOperator = registry.isOwnerOrOperator(args.parentHash, msg.sender);
+        if (parentConfig.accessType == AccessType.LOCKED && !isOwnerOrOperator)
+            revert ParentLockedOrDoesntExist(args.parentHash);
+
+        if (parentConfig.accessType == AccessType.MINTLIST) {
+            if (
+                !mintlist[args.parentHash]
+            .list
+            [mintlist[args.parentHash].ownerIndex]
+            [msg.sender]
```
JamesEarle, line 164: indent these for clarity

init-env.ts
```diff
@@ -0,0 +1,3 @@
+import { setDefaultEnvironment } from "./src/environment/set-env";
+
+setDefaultEnvironment();
```
JamesEarle, line 3: why not just have the hh config import this directly?

contracts/cross-chain/ZNSZChainPortal.sol
Whytecrowe: the downside here is that every time this enum is extended, we need to upgrade all contracts that use it

contracts/cross-chain/mock/PolygonZkEVMBridgeV2Mock.sol
Whytecrowe: This is a mock of a Polygon Bridge contract. the names here are Polygon names.

hardhat.config.ts
Whytecrowe: These are public links that users would use. is there a reason for make them env vars?

contracts/registrar/ZNSRootRegistrarBase.sol
Whytecrowe: Agreed. no idea how this changed tbh :)

init-env.ts
Whytecrowe: you mean dotenv? that's importing different things. or do you mean add this call to the HH config instead of this file?

src/environment/set-env.ts
Whytecrowe: not sure what you mean. we use the defaults from the file on the line above this, then this one overrides all the default from the env file if present.
env vars from .env file are loaded whenever this line is called. or am I misunderstanding?

src/environment/validate.ts
Whytecrowe: this is just basic readability to me rather then thinking about line length. here I am separating function param and def from the function code on the next line. we can split here or not split, it's not a big deal, it was just more readable to me this way, so I decided to do it.
Also, shorter lines make it easier to read code on smaller screens.

contracts/registrar/ZNSRootRegistrarBase.sol
```diff
@@ -27,17 +29,17 @@ import { ZeroAddressPassed, DomainAlreadyExists } from "../utils/CommonErrors.so
...
     ARegistryWired,
-    IZNSRootRegistrar {
+    IZNSRootRegistrarBase {
     using StringUtils for string;
 
     IZNSPricer public rootPricer;
     IZNSTreasury public treasury;
     IZNSDomainToken public domainToken;
-    IZNSSubRegistrar public subRegistrar;
+    IZNSSubRegistrarTrunk public subRegistrar;
```
Whytecrowe, line 42: why is this a Trunk contract? is this a mistake?..

PR #119, opened 2024-09-23
Modify transfer in DomainToken

JamesEarle:
Modify the original `transferFrom` method to also include updating the registry with the owner update as well, effectively tying the split ownership together in all standard transfers such as sales on OpenSea

Introduce a new non-standard transfer `transferTokenFrom` that behaves similarly to the original `transferFrom` where *only* the token is transferred.


contracts/access/ZNSAccessController.sol
```diff
@@ -34,6 +34,8 @@ contract ZNSAccessController is AccessControl, ZNSRoles, IZNSAccessController {
         _setRoleAdmin(GOVERNOR_ROLE, GOVERNOR_ROLE);
         // all of the admins control registrar
         _setRoleAdmin(REGISTRAR_ROLE, ADMIN_ROLE);
+        // all of the admins control domain token
+        _setRoleAdmin(DOMAIN_TOKEN_ROLE, ADMIN_ROLE);
```
Whytecrowe, line 38: hmm.... this will require us to upgrade every single contract in ZNS for this change to work? is it possible to do this in a less invasive way?

contracts/access/ZNSRoles.sol
```diff
@@ -22,6 +22,11 @@ abstract contract ZNSRoles {
      */
     bytes32 public constant REGISTRAR_ROLE = keccak256("REGISTRAR_ROLE");
 
+    /**
+     * @notice This role is here specifically for the ZNSDomainToken.sol contract
+     */
+    bytes32 public constant DOMAIN_TOKEN_ROLE = keccak256("DOMAIN_TOKEN_ROLE");
```
Whytecrowe, line 28: can't we just use Registrar role for this? this way we don't need to upgrade all contracts just for this string...
thoughts?

contracts/token/IZNSDomainToken.sol
```diff
@@ -55,11 +58,15 @@ interface IZNSDomainToken is IERC2981, IERC721 {
 
     function setDefaultRoyalty(address receiver, uint96 royaltyFraction) external;
 
+    function transferTokenFrom(address from, address to, uint256 tokenId) external;
```
Whytecrowe: maybe make a more clear name?...
idk... "changeTokenOwner()` or something... "transferTokenFrom" is kind of a good name, but may be a bit confusing?

contracts/token/ZNSDomainToken.sol
```diff
@@ -174,6 +190,37 @@ contract ZNSDomainToken is
...
+    /**
+     * @notice Override the standard transferFrom function to update the owner for both the `registry` and `token`
+     * 
+     * @dev See {IERC721-transferFrom}
+     */
+    function transferFrom(
+        address from,
+        address to,
+        uint256 tokenId
+    ) public override(ERC721Upgradeable, IERC721) {
```
Whytecrowe, line 213: hm... what about all other transfer methods? there used to be 3 at least. those will not do what we want.

contracts/access/ZNSRoles.sol
Whytecrowe: actually... we may have to upgrade all contracts anyway since they all have custom errors now...

JamesEarle: It's sort of a decision on the "proper" way this should be. It would definitely be the same effect to just assign the `DomainToken` contract the `REGISTRAR_ROLE` but it isn't a registrar, so it felt improper. If we have to upgrade regardless it might make sense to leave it. Let me know what you think

contracts/token/IZNSDomainToken.sol
JamesEarle: I felt making it as close to the regular `transferFrom` was best, so `transferTokenFrom` made sense because it was transferring *just* the token, but maybe `setTokenOwner` or `updateTokenOwner` to match verbiage from the registry would be better

contracts/token/ZNSDomainToken.sol
JamesEarle: Messaged you about this, but yes because the other two public `safeTransferFrom` both route through this function it's fine to just override this one.

Whytecrowe (changes requested):
looks good! just need to make sure all possible ERC721 transfer methods are covered here

PR #120, opened 2024-11-27
ZDC Update for Multi Chain [Analyze certain changes for ZNS v1.5 Not Cross chain!]

Whytecrowe:
`rc/multi-zns-main` or `multi-zns-proto` (after merge with the former) needs to be merged into this branch to update the code with all the previous changes !!!


src/deploy/campaign/get-config.ts
```diff
@@ -68,10 +68,20 @@ export const getConfig = async ({
   governors ?: Array<string>;
   admins ?: Array<string>;
   zeroVaultAddress ?: string;
-  env ?: string;
-}) : Promise<IZNSCampaignConfig<SignerWithAddress | Wallet>> => {
+  env ?: TEnvironment;
+}) : Promise<IZNSCampaignConfig> => {
+  // Prioritize reading from the env variable first, and only then fallback to the param
+  let envLevel = process.env.ENV_LEVEL as TEnvironment;
```
JamesEarle, line 74: Could simplify this and any other null checks with `const envLevel = env ?? process.env.ENV_LEVEL;` or the opposite order.

Also this seems a bit odd because it will read `ENV_LEVEL` but it will always fallback to `env` if specified, but from your comment it sounds like it should only fallback to `env` if the env var wasn't found in the system with `process.env.ENV_LEVEL`

src/deploy/campaign/get-config.ts
```diff
@@ -93,7 +103,9 @@ export const getConfig = async ({
   }
 
   // Domain Token Values
-  const royaltyReceiver = process.env.ENV_LEVEL !== "dev" ? process.env.ROYALTY_RECEIVER! : zeroVaultAddressConf;
+  const royaltyReceiver = process.env.ENV_LEVEL !== EnvironmentLevels.dev
```
JamesEarle, line 106: we could do `else if` above and it would feel cleaner than checking twice and separately

src/deploy/campaign/get-config.ts
```diff
@@ -82,7 +92,7 @@ export const getConfig = async ({
 
   let zeroVaultAddressConf : string;
```
JamesEarle, line 93: small nitpick, but what is the `Conf` suffix adding here? Makes me think this is a config, but it's just a single value. I would just use `zeroVaultAddress`

src/deploy/campaign/get-config.ts
```diff
@@ -219,7 +225,10 @@ const validatePrice = (config : ICurvePriceConfig) => {
   return (priceB <= priceA);
 };
 
-export const buildCrosschainConfig = () : TZNSCrossConfig => {
+export const buildCrosschainConfig = (env : TEnvironment) : {
+  crosschain : TZNSCrossConfig;
+  srcChainName : TSupportedChain;
+} => {
   const srcChainName = process.env.SRC_CHAIN_NAME as TSupportedChain;
```
JamesEarle, line 232: This won't actually cause an error if `proess.env.SRC_CHAIN_NAME` is not one of the one's included in `TSupportedChain`, the `as` in TypeScript is basically telling the transpiler "even if this isn't what I type it as, don't complain because I know what I'm doing"

It will be fine though because your `switch` below checks for exact values

src/deploy/campaign/get-config.ts
```diff
@@ -275,5 +305,5 @@ export const buildCrosschainConfig = () : TZNSCrossConfig => {
     throw new Error(`Unsupported chain: ${srcChainName}!`);
   }
 
-  return crossConfig;
+  return { crosschain: crossConfig, srcChainName };
```
JamesEarle, line 308: why not just use the name of the variable you're returning when you declare it above for simplicity? or change the return variable name to match? seems like a needless thing to have one prop named the return value and one not

src/deploy/campaign/types.ts
```diff
@@ -39,13 +36,14 @@ export interface IZNSEthCrossConfig extends IZNSBaseCrossConfig {
 
 export interface IZNSZChainCrossConfig extends IZNSBaseCrossConfig {
   srcZnsPortal : string;
+  ethAdmin ?: IZNSSigner;
 }
 
 export type TZNSCrossConfig = IZNSEthCrossConfig | IZNSZChainCrossConfig;
```
JamesEarle, line 42: small nitpick, but it feels redundant to start every custom type with `T`, we know they are a type by the way we reference it

src/deploy/campaign/get-config.ts
Whytecrowe: that's your code and comment :)

contracts/cross-chain/ZNSZChainPortal.sol
```diff
@@ -71,8 +71,12 @@ contract ZNSZChainPortal is UUPSUpgradeable, AAccessControlled, IZNSZChainPortal
         bytes32 parentHash,
         string calldata label,
         string calldata tokenURI
+    // TODO multi: figure this out !!!
 //      bool forceUpdateGlobalExitRoot  - do we need to pass this ???
     ) external override {
+        // TODO multi: can we optimize this somehow, so we don't have to do it every time ??
```
JamesEarle, line 77: Why are the other destination values set in the initializer but not this one? If we set this in the initializer we could be sure it's non-zero, as it also reverts when address(0) is given in the setter function below

src/deploy/missions/contracts/cross-chain/zk-evm-bridge.ts
```diff
@@ -23,8 +22,8 @@ IZNSContracts
 
   constructor (args : IDeployMissionArgs<
```
JamesEarle, line 23: It looks like this type is only used a single time. It would be simpler to just make the type you need instead of creating a single-use generic type then typing it with the expected values anyways

PR #124, opened 2025-02-14
[V2 MAIN BRANCH] ZNS v2 for Z Chain [WIP]

Whytecrowe:


contracts/price/ZNSCurvePricer.sol
```diff
@@ -250,19 +293,26 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
      */
     function setPrecisionMultiplier(
         bytes32 domainHash,
         uint256 multiplier
     ) public override onlyOwnerOrOperator(domainHash) {
-        require(multiplier != 0, "ZNSCurvePricer: precisionMultiplier cannot be 0");
-        require(multiplier <= 10**18, "ZNSCurvePricer: precisionMultiplier cannot be greater than 10^18");
-        priceConfigs[domainHash].precisionMultiplier = multiplier;
-
+        _setPrecisionMultiplier(domainHash, multiplier);
```
JamesEarle: If all the setters call to internal functions that both validate and set the variable they should be named the same, `_validateSet[VAR_NAME]` like above with `_validateSetMaxLength(...)`

PR #125, opened 2025-02-20
[v1.5: 5] Update meow implementation

Whytecrowe:


contracts/token/mocks/ERC20Mock.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+/* solhint-disable */
+pragma solidity 0.8.26;
+
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+
+contract ERC20Mock is ERC20 {
```
JamesEarle, line 8: small nitpick, but `MockERC20` is clearer and consistent with the way we always name it

test/helpers/deploy/deploy-zns.ts
```diff
@@ -171,45 +171,31 @@ export const deployDomainToken = async (
...
-      meowTokenSymbol,
-    ],
-    {
-      kind: "transparent",
-    }
-  ) as unknown as MeowTokenMock;
+  const meowToken = await factory.deploy(
+    meowTokenName,
+    meowTokenSymbol,
+  ) as unknown as ERC20Mock;
```
JamesEarle, line 179: why `as unknown` here before `as ERC20Mock`? Coming from the `ERC20Mock__factory` above the type shouldn't need `unknown` first

JamesEarle (approved):
LGTM

PR #126, opened 2025-02-25
[v1.5: 6] Environment structure and zDC Update 

Whytecrowe:


src/deploy/missions/contracts/meow-token/meow-token.ts
```diff
@@ -64,7 +64,7 @@ IZNSContracts
       this.logger.debug(`Writing ${this.contractName} to DB...`);
 
       const factory = new ZToken__factory(this.config.deployAdmin);
-      const baseContract = factory.attach(this.config.stakingTokenAddress);
+      const baseContract = factory.attach(this.config.stakingTokenAddress as string);
```
JamesEarle, line 67: This is odd, can TypeScript not see the value as a string already?

Whytecrowe: it could be undefined

src/environment/types.ts
```diff
@@ -0,0 +1,47 @@
+import { TEnvironment } from "@zero-tech/zdc";
+
+
+export interface IZNSEnvironment {
```
JamesEarle, line 4: Although it is an object and ordering doesn't matter for functions that use this, it will still be easier to read if all the nullable properties come after non-nullable ones as opposed to an arbitrary order

JamesEarle (approved):
LGTM

PR #130, opened 2025-03-26
[v1.0] Upgrade to pausable contracts and system lock

Whytecrowe:


contracts/zns-pausable/registrar/ZNSRootRegistrarPausable.sol
```diff
@@ -0,0 +1,453 @@
...
+     *      + `registrant`: The address of the user who is registering the domain
+     *      + `price`: The determined price for the domain to be registered based on parent rules
+     *      + `stakeFee`: The determined stake fee for the domain to be registered (only for PaymentType.STAKE!)
+     *      + `domainAddress`: The address to which the domain will be resolved to
+     *      + `tokenURI`: The tokenURI for the domain to be registered
+     *      + `isStakePayment`: A flag for whether the payment is a stake payment or not
+    */
+    function coreRegister(
+        CoreRegisterArgs memory args
+    ) external override onlyRegistrar {
```
JamesEarle: Should this be `whenNotPaused` as well? This `ZNSRootRegistrar` and `ZNSSubRegistrar` could hypothetically be out of sync and making this function also guarded with `whenNotPaused` would disallow subdomain registration even in the case where the subdomain registrar is not paused yet

contracts/zns-pausable/IZNSPausable.sol
JamesEarle, line 1: Should this also include `function paused` ?

contracts/zns-pausable/registrar/ZNSRootRegistrarPausable.sol
```diff
@@ -0,0 +1,453 @@
...
+        _coreRegister(
+            args
+        );
+    }
+
+    /**
+     * @dev Returns true if the contract is paused, and false otherwise.
+     */
+    function paused() public view virtual returns (bool) {
```
JamesEarle: virtual?

contracts/zns-pausable/registry/ZNSRegistryPausable.sol
```diff
@@ -0,0 +1,353 @@
...
+     */
+    function isOwnerOrOperator(
+        bytes32 domainHash,
+        address candidate
+    ) public view override returns (bool) {
+        address owner = records[domainHash].owner;
+        return candidate == owner || operators[owner][candidate];
+    }
+
+    function paused() public view returns (bool) {
```
JamesEarle: This should be grouped with the other functions from pausable to be consistent with the other contracts. 

Also, no `virtual` on this one but it is present in other places?

contracts/zns-pausable/token/ZNSDomainTokenPausable.sol
```diff
@@ -0,0 +1,289 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.18;
+
+import { ERC721Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";
+import { IERC721Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC721/IERC721Upgradeable.sol";
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import { ERC2981Upgradeable } from "@openzeppelin/contracts-upgradeable/token/common/ERC2981Upgradeable.sol";
+import { ERC721URIStorageUpgradeable }
+    from "@openzeppelin/contracts-upgradeable/token/ERC721/extensions/ERC721URIStorageUpgradeable.sol";
```
JamesEarle: This indentation is whacky,  but I assume it's here because it breaks line length in linting

Consider this instead

```solidity
import {
    ERC721URIStorageUpgradeable 
} from "@openzeppelin/contracts-upgradeable/token/ERC721/extensions/ERC721URIStorageUpgradeable.sol";
```

contracts/zns-pausable/token/ZNSDomainTokenPausable.sol
```diff
@@ -0,0 +1,289 @@
...
+        uint256 tokenId,
+        address receiver,
+        uint96 royaltyFraction
+    ) external override onlyAdmin {
+        _setTokenRoyalty(tokenId, receiver, royaltyFraction);
+
+        emit TokenRoyaltySet(tokenId, royaltyFraction);
+    }
+
+    function paused() public view returns (bool) {
```
JamesEarle: Also be sure to add natspec to each instance where this is referenced

hardhat.config.ts
JamesEarle, line 1: Looks like a line ending issue again?

package.json
JamesEarle, line 1: same here with EOL characters

src/upgrade/db.ts
```diff
@@ -0,0 +1,97 @@
+import * as hre from "hardhat";
+import { MongoDBAdapter } from "../deploy/db/mongo-adapter/mongo-adapter";
+import { VERSION_TYPES } from "../deploy/db/mongo-adapter/constants";
+import { getLogger } from "../deploy/logger/create-logger";
+import { znsNames } from "../deploy/missions/contracts/names";
+import { getContractDataForUpgrade } from "./upgrade";
+
+
+export const updateAllContractsInDbAndVerify = async ({
```
JamesEarle: I think `updateAndVerify`  or `updateAndVerifyAll` would be enough for the function name considering it is within the context already of something that receives a db adapter and so it can be inferred that it will be operating on the DB

src/upgrade/db.ts
```diff
@@ -0,0 +1,97 @@
...
+  dbAdapter,
+} : {
+  dbAdapter : MongoDBAdapter;
+}) => {
```
JamesEarle: using this function parameter syntax with `{ ...args } : { ...argTypes }` feeels a bit clunky when only a single parameter is used, the whole "parameter ordering" argument can't apply here either

src/upgrade/db.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  const newDbVersion = Date.now().toString();
+  const newContractsVersion = dbAdapter.getContractsVersionFromFile();
+  logger.info(
+    `Updating DB "${dbAdapter.dbName}" with new version: ${newDbVersion} and contracts version: ${newContractsVersion}`
+  );
+
+  for (const { contractName, address } of contractData) {
+    let implAddress : string | null;
+    if (contractName === znsNames.accessController.contract) {
```
JamesEarle, line 30: Is there a better way we can identify an upgradeable contract instead of just having it check this one specific hardcoded address we know isn't an upgradeable contract? You already have hardhat here, it would be easy to check the impl address from there. It would throw an exception if you a provide a non-upgradeable contract, so maybe `try / catch` it to be able to proceed without error

src/upgrade/db.ts
```diff
@@ -0,0 +1,97 @@
...
+  await dbAdapter.versions.insertOne({
+    type: VERSION_TYPES.upgraded,
+    dbVersion: newDbVersion,
+    contractsVersion: newContractsVersion,
+  });
+
+  logger.info("DB update finished successfully.");
+};
+
+export const updateContractInDb = async ({
```
JamesEarle, line 63: similar to above I feel like the suffix `...InDb` is unnecessary here because the file itself is `db.ts` can takes in a `dbAdapter` for this upgrade module

src/upgrade/db.ts
```diff
@@ -0,0 +1,97 @@
...
+  contractName : string;
+  proxyAddress : string;
+  implAddress : string | null;
+  newDbVersion : string;
+}) => {
+  const artifactName = contractName === znsNames.accessController.contract
+    || contractName === znsNames.meowToken.contract
+    || contractName === znsNames.meowToken.contractMock
+    ? contractName
+    : `${contractName}Pausable`;
```
JamesEarle, line 80: is it not possible for an incoming contract name to already include `Pausable` as a suffix?

src/deploy/db/mongo-adapter/get-adapter.ts
JamesEarle, line 1: EOL

src/upgrade/db.ts
Whytecrowe: I like to specify we are updating DB there. It's harder to tell what it is when just looking at the name of the function. having DB doesn't hurt, but make it more clear.

src/upgrade/db.ts
Whytecrowe: I prefer to keep it more specified here. If we use try/catch here and certain contract that we do want fails for any other reason, it will not update the implementation and not raise any error, so we will miss a valid impl in the DB.
This is a safer version, since we do know for sure that accessController is the only contract we are not upgrading here, and if other ones fails for some reason, we will know.

src/upgrade/db.ts
Whytecrowe: technically, not possible, since we are using a constant map we have that specify base contract names without Pausable

contracts/zns-pausable/price/ZNSCurvePricerPausable.sol
JamesEarle, line 1: After my changes are merged to add stateless pricer modules, we won't need to consider adding pausable to this or the FixedPricer

Whytecrowe: this is for the different version.

PR #131, opened 2025-03-29
[v1.5] Bulk domain minting

MichaelKorchagin:


contracts/registrar/IZNSRootRegistrar.sol
```diff
@@ -50,6 +50,14 @@ interface IZNSRootRegistrar is IDistributionConfig {
 
     error InvalidOwnerOfEnumValue(OwnerOf value);
 
+    struct RootDomainRegistration {
```
Whytecrowe: RootDomainRegistrationParams

contracts/registrar/IZNSSubRegistrar.sol
```diff
@@ -22,6 +22,15 @@ interface IZNSSubRegistrar is IDistributionConfig {
      */
     error SenderNotApprovedForPurchase(bytes32 parentHash, address sender);
 
+    struct SubDomainRegistration {
```
Whytecrowe: same here. add Params, so we know it's where it's used

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -130,6 +130,38 @@ contract ZNSRootRegistrar is
         return domainHash;
     }
 
+    /**
+     * @notice This function allows registering multiple root domains in a single transaction.
+     * It iterates through an array of `RootDomainRegistration` objects, registering each domain
+     * by calling the `registerRootDomain` function for each entry.
+     * @dev This function reduces the number of transactions required to register multiple domains,
+     * saving gas and improving efficiency. Each domain registration is processed sequentially.
```
Whytecrowe, line 138: Please add information on how domain data can be passed and how this function works with domains of the same path or domains of different parents.
Also, add all of the props of the struct, specifically explain how the value of `parentHash` prop influences the behaviour. E.g. if `parentHash` under index other than 0 will signal that this domain is the child of the one under idx-1 and so on...

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -85,16 +85,16 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
-    ) external override returns (bytes32) {
+        string memory tokenURI,
+        DistributionConfig memory distrConfig,
+        PaymentConfig memory paymentConfig
+    ) public override returns (bytes32) {
         // Confirms string values are only [a-z0-9-]
         label.validate();
 
-        bytes32 domainHash = hashWithParent(parentHash, label);
+        bytes32 domainHash = hashWithParent(parentHash, label);
```
Whytecrowe: seems like extra whitespace was added here.

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -155,12 +155,65 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+     * - This pattern can continue for deeper levels of nesting.
+     * 
+     * @param subRegistrations An array of `SubDomainRegistration` structs, each containing:
+     *      + `parentHash`: The hash of the parent domain under which the subdomain is being registered.
+     *                     If set to `0x000...`, the parent will be the previously registered subdomain.
+     *      + `label`: The label of the subdomain to register (e.g., in `0://parent.child`, the label is `child`).
+     *      + `domainAddress`: The address to associate with the subdomain in the resolver.
+     *      + `tokenURI`: The URI to assign to the subdomain token.
+     *      + `distributionConfig`: The distribution configuration for the subdomain.
+     *      + `paymentConfig`: The payment configuration for the subdomain.
```
Whytecrowe, line 176: add missing props of the struct here and specify that the initial `parentHash` can NOT be 0x0, but all subsequent ones will be registered as children of the previous if `parentHash` is 0x0.

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -155,12 +155,65 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+    ) external override returns (bytes32[] memory) {
+
+        bytes32[] memory domainHashes = new bytes32[](subRegistrations.length);
+    
+        for (uint256 i = 0; i < subRegistrations.length; i++) {
+            if (subRegistrations[i].parentHash == bytes32(0) && i > 0) {
+                subRegistrations[i].parentHash = domainHashes[i - 1];
+
+            } else if (subRegistrations[i].parentHash == bytes32(0)) {
+                revert ZeroAddressPassed();
```
Whytecrowe: wrong error. if we do add a specific check like this here, we need a specific error then too so that user knows what happened

hardhat.config.ts
```diff
@@ -54,26 +54,26 @@ const config : HardhatUserConfig = {
...
+    //   "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol": {
+    //     version: "0.8.20",
+    //     settings: {
+    //       optimizer: {
+    //         enabled: true,
+    //         runs: 20000,
+    //       },
+    //     },
+    //   },
+    // },
```
Whytecrowe: why is this commented out?!

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,6 +127,182 @@ describe("ZNSSubRegistrar", () => {
...
+    it("Should #registerMultipleSubdomains and the event must be triggered", async () => {
+      const registrations = [];
+
+      for (let i = 0; i < 5; i++) {
+        const isOdd = i % 2 !== 0;
+
+        const subdomainObj = {
+          parentHash: rootHash,
+          label: `subdomain${i + 1}`,
+          domainAddress: admin.address,
```
Whytecrowe, line 139: all of these are registered by the same account, which means all the logic with payments and pricing is circumvented. we need a case where registrations are for different owners as well

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,6 +127,182 @@ describe("ZNSSubRegistrar", () => {
...
+      // how many nested domains (0://root.sub1.sub2.sub3....)
+      const domainLevels = 15;
+
+      for (let i = 0; i < domainLevels; i++) {
+        const isOdd = i % 2 !== 0;
+
+        const subdomainObj = {
+          parentHash: rootHash,
+          label: `sub${i + 1}`,
+          domainAddress: lvl3SubOwner.address,
```
Whytecrowe, line 194: same here. all are registered by the same account. we also need to check diff accounts

contracts/registrar/IZNSSubRegistrar.sol
Whytecrowe: also, Subdomain, not SubDomain

Whytecrowe (changes requested):
address comments

contracts/registrar/ZNSSubRegistrar.sol
MichaelKorchagin: I understand that it may be long and difficult for the user to read, otherwise this comment does not make sense to me.
1. All the props have been added to the description at L174-L181. I do not think we need to describe `distributionConfig` and `paymentConfig` twice, since these are parameters of another function, where they are already described. Also, they are explained in the interface, according to the struct
2. I specified when `parentHash` could be 0x0 and when it couldn't at L165-L172. Duplicated at L175-L176 in params section

What am I missing?

hardhat.config.ts
MichaelKorchagin: The compiler version for the following files is fixed as "^0.8.22". Uncommented and changed to 0.8.22.

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -94,7 +94,7 @@ contract ZNSRootRegistrar is
         string calldata tokenURI,
         DistributionConfig calldata distributionConfig,
         PaymentConfig calldata paymentConfig
```
Whytecrowe: code should work the same way everywhere. If you turned params into a struct, params for this function should be the same struct as well.

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -130,6 +130,38 @@ contract ZNSRootRegistrar is
...
+     *      + `paymentConfig`: The payment configuration for the domain (optional).
+     * @return domainHashes An array of `bytes32` hashes representing the registered domains.
+     */
+    function registerMultipleRootDomains(
+        RootDomainRegistrationParams[] calldata registrations
+    ) external override returns (bytes32[] memory) {
+        bytes32[] memory domainHashes = new bytes32[](registrations.length);
+
+        for (uint256 i = 0; i < registrations.length; i++) {
+            domainHashes[i] = registerRootDomain(
```
Whytecrowe: after you change this function params to a struct, then you can just easily pass the struct under a specific index here, reducing code.

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -85,12 +85,12 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+        string memory label,
         address domainAddress,
-        string calldata tokenURI,
-        DistributionConfig calldata distrConfig,
-        PaymentConfig calldata paymentConfig
-    ) external override returns (bytes32) {
+        string memory tokenURI,
+        DistributionConfig memory distrConfig,
+        PaymentConfig memory paymentConfig
+    ) public override returns (bytes32) {
```
Whytecrowe, line 88: same here. this functions params should be the same struct as in the multi function.

test/ZNSRootRegistrar.test.ts
```diff
@@ -103,6 +103,73 @@ describe("ZNSRootRegistrar", () => {
...
+      paymentConfig: {
+        token: await zns.meowToken.getAddress(),
+        beneficiary: admin.address,
+      },
+    };
+
+    // Attempt to register the same domain again
+    await expect(
+      zns.rootRegistrar.connect(user).registerMultipleRootDomains([domainObj, domainObj])
+    ).to.be.revertedWithCustomError(zns.rootRegistrar, "DomainAlreadyExists");
```
Whytecrowe: do not use hardcoded errors. almost all errors are present as constanst in errors.ts file. if error is not present - add it there, then use the constant

test/ZNSRootRegistrar.test.ts
```diff
@@ -103,6 +103,73 @@ describe("ZNSRootRegistrar", () => {
...
+  });
+
```
Whytecrowe, line 172: check actual domain registration as well:
1. registry record set properly
2. token minted properly to the proper address

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,6 +127,182 @@ describe("ZNSSubRegistrar", () => {
...
+              subdomainObj.label
+            )
+          );
+        } else {
+          parentHashes.push(
+            await zns.subRegistrar.hashWithParent(
+              parentHashes[i - 1],
+              subdomainObj.label
+            )
+          );
```
Whytecrowe: can be written with less code:
```typescript
parentHashes.push(
            await zns.subRegistrar.hashWithParent(
              i === 0 ? rootHash : parentHashes[i - 1],
              subdomainObj.label
            )
          );
```
and no large If is needed with the same call written twice.

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,6 +127,182 @@ describe("ZNSSubRegistrar", () => {
...
+      };
+
+      // Add allowance
+      await zns.meowToken.connect(lvl2SubOwner).approve(await zns.treasury.getAddress(), ethers.MaxUint256);
+
+      await expect(
+        zns.subRegistrar.connect(lvl2SubOwner).registerMultipleSubdomains([subdomainObj])
+      ).to.be.revertedWithCustomError(zns.subRegistrar, ZERO_ADDRESS_ERR);
+    });
+
```
Whytecrowe, line 438: needs more tests.
what about combinations? when you register some nested ones and some non-nested?
I register 10 subs, first 4 are nested, then 2 non-nested with specific parent hashes, then 2 nested and the rest non-nested.
this also needs to be properly validated in the system, not just reading the event.
Was each sub created under the proper parent? we can tell this by validating the hash produced for each with a reference hash where we hash specific ones with their proper parents.

Whytecrowe (changes requested):
please, address comments

test/ZNSRootRegistrar.test.ts
MichaelKorchagin: It's already checked in the tests:
1. "Sets the correct data in Registry".
2. "Stakes and saves the correct amount and token, takes the correct fee and sends fee to Zero Vault".

Whytecrowe: it is checked for a completely different case. we need to make sure all the domain registrations from your specific multi registration use case

contracts/registrar/IZNSRootRegistrar.sol
```diff
@@ -134,13 +143,13 @@ interface IZNSRootRegistrar is IDistributionConfig {
...
     function registerRootDomain(
-        string calldata name,
-        address domainAddress,
-        string calldata tokenURI,
-        DistributionConfig calldata distributionConfig,
-        PaymentConfig calldata paymentConfig
+        RootDomainRegistrationParams calldata registration
     ) external returns (bytes32);
 
+    function registerMultipleRootDomains(
```
JamesEarle: Ideally it would have the same name as the singular function just with `Bulk` or similar at the end, so `registerRootDomainBulk( ... )`

contracts/registrar/IZNSSubRegistrar.sol
```diff
@@ -91,18 +92,12 @@ interface IZNSSubRegistrar is IDistributionConfig {
...
-        DistributionConfig calldata configForSubdomains,
-        PaymentConfig calldata paymentConfig
+        SubdomainRegistrationParams calldata registration
     ) external returns (bytes32);
 
-    function hashWithParent(
-        bytes32 parentHash,
-        string calldata label
-    ) external pure returns (bytes32);
+    function registerMultipleSubdomains(
```
JamesEarle: same with naming as other comment, `registerSubdomainBulk` or similar

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -77,59 +77,82 @@ contract ZNSRootRegistrar is
...
+     *      + `distributionConfig`: The distribution configuration for the domain (optional).
+     *      + `paymentConfig`: The payment configuration for the domain (optional).
+     * @return domainHashes An array of `bytes32` hashes representing the registered domains.
+     */
+    function registerMultipleRootDomains(
+        RootDomainRegistrationParams[] calldata registrations
+    ) external override returns (bytes32[] memory) {
+        bytes32[] memory domainHashes = new bytes32[](registrations.length);
+
+        for (uint256 i = 0; i < registrations.length; i++) {
```
JamesEarle: minor gas improvement is made by incrementing the counter outside of the loop in an unchecked block, and for such a heavy operation that will matter a lot

```solidity
uint256 i = 0;
for(i; i < registrations.length;) {
  ...
  unchecked {
    ++i;  
  }
}
```

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -148,19 +143,65 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+     *      + `paymentConfig`: The payment configuration for the subdomain.
+     * @return domainHashes An array of `bytes32` hashes representing the registered subdomains.
+     */
+    function registerMultipleSubdomains(
+        SubdomainRegistrationParams[] memory subRegistrations
+    ) external override returns (bytes32[] memory) {
+
+        bytes32[] memory domainHashes = new bytes32[](subRegistrations.length);
+
+        for (uint256 i = 0; i < subRegistrations.length; i++) {
```
JamesEarle: same as other comment for gas savings with how we increment the counter

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -148,19 +143,65 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+            if (subRegistrations[i].parentHash == bytes32(0) && i > 0) {
+                subRegistrations[i].parentHash = domainHashes[i - 1];
```
JamesEarle: this is weird, do we know with certainty the ordering of incoming registration params?

say first is root, then 10 more domains get minted where parent hash is root
second domain when i = 1, parenthash is 0x0, so now parenthash is set to the previous domain hash (correct)
third domain when i = 2, parentHash is 0x0 again, now we set parentHash to ACTUALLY be the hash of the last domain we minted, which would be incorrect

Am I interpreting this right? why do we do this?

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -148,19 +143,65 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+
+            } else if (subRegistrations[i].parentHash == bytes32(0)) {
```
JamesEarle: Try to always add code that reverts as early as possible in the function to save gas, having the opposite of the above check would make this easier and then switch the order

hardhat.config.ts
```diff
@@ -91,29 +91,29 @@ const config : HardhatUserConfig = {
...
-      // ],
-      // // Must have to avoid instead failing as `invalid length for result data` error
-      // throwOnCallFailures: false, // not sure if this even works
-    },
-    devnet: {
-      // Add current URL that you spawned if not using automated spawning
-      url: `${process.env.DEVNET_RPC_URL}`,
-      chainId: 1,
-    },
+    // mainnet: {
```
JamesEarle, line 94: is there a need to comment out these? cant we just leave them and not use them?

test/ZNSRootRegistrar.test.ts
```diff
@@ -103,6 +103,73 @@ describe("ZNSRootRegistrar", () => {
     await mongoAdapter.dropDB();
   });
 
+  it("Should register an array of domains", async () => {
+    const registrations = [];
+
+    for (let i = 0; i < 5; i++) {
+      const isOdd = i % 2 !== 0;
```
JamesEarle, line 110: doesnt really make sense to store a simple check like this in a variable just to check it again below, we can just directly read it 

```typescript
...
paymentType: i % 2 === 0 ? PaymentType.STAKE : PaymentType.DIRECT,
```

test/ZNSRootRegistrar.test.ts
```diff
@@ -103,6 +103,73 @@ describe("ZNSRootRegistrar", () => {
     await mongoAdapter.dropDB();
   });
 
+  it("Should register an array of domains", async () => {
+    const registrations = [];
+
+    for (let i = 0; i < 5; i++) {
+      const isOdd = i % 2 !== 0;
+
+      const domainObj = {
```
JamesEarle: create a type for this object if we do not have one already so that in the future should we need to change anything or redo similar work, it shows an error here when the type itself is modified

test/ZNSRootRegistrar.test.ts
```diff
@@ -103,6 +103,73 @@ describe("ZNSRootRegistrar", () => {
...
+      expect(args[1]).to.eq(hashDomainLabel(domain.name)); // domainHash
+      expect(args[2]).to.eq(domain.name); // label
+      expect(args[4]).to.eq(domain.tokenURI); // tokenURI
+      expect(args[5]).to.eq(user.address); // registrant
+      expect(args[6]).to.eq(domain.domainAddress); // domainAddress
+    }
+  });
+
+  it("Should revert when register the same domain twice using #registerMultipleRootDomains", async () => {
+    const domainObj = {
```
JamesEarle, line 152: same as above, create and use a type if one doesnt already exist. This will make sure that all these tests use the same thing and future proof it

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -148,19 +143,65 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+                revert ZeroParentHash(subRegistrations[i].label);
```
JamesEarle: But what if the subdomain's parent is the root domain?

test/helpers/events.ts
```diff
@@ -12,7 +12,7 @@ export const getDomainRegisteredEvents = async ({
   registrant = undefined,
   blockRange = 50,
 } : {
-  zns : IZNSContractsLocal;
+  zns : IZNSContractsLocal | IZNSContracts;
```
JamesEarle, line 15: ?

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,22 +136,423 @@ describe("ZNSSubRegistrar", () => {
...
+      // eslint-disable-next-line @typescript-eslint/prefer-for-of
+      for (let i = 0; i < logs.length; i++) {
+        const subdomain = registrations[i];
+
+        // "DomainRegistered" event log
+        const args = logs[i].args;
+
+        expect(args[0]).to.eq(rootHash); // parentHash
+        expect(args[2]).to.eq(subdomain.label); // label
```
JamesEarle: should we also have `expect(args[1])...` ?

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,22 +136,423 @@ describe("ZNSSubRegistrar", () => {
...
+    });
+
+    it("Should register multiple NESTED subdomains using #registerMultipleSubdomains", async () => {
+      const registrations = [];
+      const parentHashes : Array<string> = [];
+
+      // how many nested domains (0://root.sub1.sub2.sub3....)
+      const domainLevels = 15;
+
+      for (let i = 0; i < domainLevels; i++) {
```
JamesEarle, line 209: Abstract this into a helper function instead of having it included in each test. A function that can create `n` configs objects and returns an array of them all, then you wont have this loop in every test

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,22 +136,423 @@ describe("ZNSSubRegistrar", () => {
...
+        expect(args[1]).to.eq(parentHashes[i]);         // domainHash current
+        expect(args[2]).to.eq(subdomain.label);         // label
+        expect(args[4]).to.eq(subdomain.tokenURI);      // tokenURI
+        expect(args[5]).to.eq(lvl3SubOwner.address);    // registrant
+        expect(args[6]).to.eq(subdomain.domainAddress); // domainAddress
+      }
+    });
+
+    it("Should revert when register the same domain twice using #registerMultipleSubdomains", async () => {
+      const subdomainObj = {
```
JamesEarle: same here as in other places. create a helper function to create any number of the domain objects you need and stop having a totally new one in each test to avoid unnecessary bloat

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,22 +136,423 @@ describe("ZNSSubRegistrar", () => {
...
+      await zns.meowToken.connect(lvl2SubOwner).approve(await zns.treasury.getAddress(), ethers.MaxUint256);
+
+      await expect(
+        zns.subRegistrar.connect(lvl2SubOwner).registerMultipleSubdomains([subdomainObj, subdomainObj])
+      ).to.be.revertedWithCustomError(zns.subRegistrar, "DomainAlreadyExists");
+    });
+
+    it("Should revert with #ZeroAddressPassed when 1st level subdomain has zerohash", async () => {
+      const subdomainObj = {
+        parentHash: ethers.ZeroHash,
```
JamesEarle, line 295: why is parent hash === root (0x0) a fail case?

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,22 +136,423 @@ describe("ZNSSubRegistrar", () => {
...
+        "after0non0nested0",
+        "after0non0nested1",
+        "non0nested0",
+        "non0nested1",
+      ];
+      const specificParentHashes = [];
+      const expectedHashes : Array<string> = [];
+
+      // 4 nested: first has parentHash, the rest have parentHash = 0 (nested)
+      for (let i = 0; i < 4; i++) {
```
JamesEarle: You have over 140 lines here dedicated to setup for the test. Here and many other places will benefit from abstraction and creating a helper function to do this part for you. please abstract all of this and below into helpers instead of generating each one manually and uniquely. This is all pre-test configuration for what you're **actually** testing.

test/ZNSSubRegistrar.test.ts
```diff
@@ -122,22 +136,423 @@ describe("ZNSSubRegistrar", () => {
...
+
+      // Add allowance
+      await zns.meowToken.connect(lvl2SubOwner).approve(await zns.treasury.getAddress(), ethers.MaxUint256);
+
+      await expect(
+        zns.subRegistrar.connect(lvl2SubOwner).registerMultipleSubdomains([subdomainObj])
+      ).to.be.revertedWithCustomError(zns.subRegistrar, ZERO_PARENTHASH_ERR);
+    });
+
+    it("Should register a mix of nested and non-nested subdomains and validates hashes", async () => {
```
JamesEarle, line 318: regarding some discussions we've had recently, it may be best to organize this into multiple smaller tests that are within a `describe` suite. This way you avoid having a single ~230 line test for quite a few different scenarios and instead have a test specific for each one. This also helps because it is more specific as to what is failing when tests are separate. Currently if this test fails there could be many potential reasons

test/ZNSRootRegistrar.test.ts
MichaelKorchagin: But I use `isOdd` 3 times to set `accessType` and `beneficiary` as well. Are you sure it's not a good way?

contracts/registrar/ZNSSubRegistrar.sol
MichaelKorchagin: We just pass its hash to the function. But it won't be 0x0

test/helpers/events.ts
MichaelKorchagin: Type IZNSContractsLocal requires zeroVaultAddress, but IZNSContracts no. And when we fill this zns var up with campaign.state.contracts in tests, there is no zeroVaultAddress field. But #getDomainRegisteredEvents requires zns to be with ZeroVaultAddress. So I decided to make zns being one of those types in this fucntion. I could make zeroVaultAddress optional in IZNSContractsLocal , but I think it's not as safe as I did.
Let me know, if I'm not right or not clear somewhere

test/ZNSSubRegistrar.test.ts
MichaelKorchagin: Because if we pass zeroHash there, it means we want to create a nested subdomain. But if this subdomain is first in the line, we don't know who will be parent of upper level.
I think the test title should be better. I'll change it

PR #132, opened 2025-04-14
[v1.5] Fix a bug of setting wrong accesscontroller

MichaelKorchagin:


contracts/access/AAccessControlled.sol
```diff
@@ -66,8 +67,12 @@ abstract contract AAccessControlled {
...
     function _setAccessController(address _accessController) internal {
-        if (_accessController == address(0)) revert ZeroAddressPassed();
-        accessController = IZNSAccessController(_accessController);
-        emit AccessControllerSet(_accessController);
+        bytes4 interfaceId = type(IZNSAccessController).interfaceId;
+        if (ERC165(_accessController).supportsInterface(interfaceId)) {
+            accessController = IZNSAccessController(_accessController);
+            emit AccessControllerSet(_accessController);
+        } else {
+            revert WrongAccessControlAddress(_accessController);
```
Whytecrowe: AccessControlLER

Whytecrowe (changes requested):
This solution doesn't actually work as expected.
One of the main requirements in this is to not have to upgrade any of the linking contracts (that inherit from AAccessControlled). Currently the problem is that we are checking the full interface compatibility of IZNSAccessController interface.
So in the future, if we change at least a single function in this contract, the compatibility check will fail in the linking contract, because the interface of a new AccessController we are setting in state will be different.

Instead, we need to rely on something that is less likely to change. Consider this:

```solidity
// Validate if `msg.sender` has the admin role in the *current* contract
if (!IZNSAccessControllerMin(accessControllerInState).hasRole(role, msg.sender)) {
    revert NotAuthorized();
}

// Similarly, validate admin alignment in the *new* contract
try IZNSAccessControllerMin(_accessControllerToSet).hasRole(role, msg.sender) returns (bool valid) {
    if (!valid) {
        revert InvalidContractA(_accessControllerToSet);
    }
} catch {
    revert InvalidContractA(_accessControllerToSet);
}
```

Here, we are validating 2 ways:
1. We are making sure that the new AccessController we are setting in this function (`_accessControllerToSet`) has `checkAdmin(address caller)` function
2. We are making sure that the new AccessController has ADMIN set to the same address (caller) as the current AccessController we have in state now.


In order to make sure this stays consistent in the future, we need to separate this function into its own interface as such:
```solidity
interface IZNSAccessControllerMin {
  function hasRole(bytes32 role, address caller) external;
}
```

and then require this interface to be ALWAYS inherited by all current and future versions of ZNSAccessController.

contracts/access/AAccessControlled.sol
```diff
@@ -1,8 +1,12 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.26;
 
+
 import { IZNSAccessController } from "./IZNSAccessController.sol";
-import { ZeroAddressPassed } from "../utils/CommonErrors.sol";
+import { ZeroAddressPassed, WrongAccessControllerAddress, NotAuthorized } from "../utils/CommonErrors.sol";
+import { ERC165 } from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import { IAccessControl } from "@openzeppelin/contracts/access/IAccessControl.sol";
+import { ZNSRoles } from "./ZNSRoles.sol";
```
Whytecrowe: redundant import

contracts/access/IZNSRoles.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.26;
+
+
+interface IZNSRoles {
+    function ADMIN_ROLE() external view returns (bytes32);
+}
```
Whytecrowe: need getters for all roles here

contracts/access/ZNSAccessController.sol
```diff
@@ -81,6 +81,19 @@ contract ZNSAccessController is AccessControl, ZNSRoles, IZNSAccessController {
...
+    }
+
+    function supportsInterface(bytes4 interfaceId)
+    public
+    view
+    virtual
+    override(AccessControl, IZNSAccessController)
+    returns (bool) {
+        return interfaceId == type(IZNSAccessController).interfaceId;
+    }
```
Whytecrowe: redundant logic

contracts/utils/CommonErrors.sol
```diff
@@ -8,6 +8,10 @@ error DomainAlreadyExists(bytes32 domainHash);
 
 error NotAuthorizedForDomain(address caller, bytes32 domainHash);
 
+error WrongAccessControllerAddress(address accessController);
+
+error NotAuthorized(address caller);
```
Whytecrowe: do we need this? we should already have access control error

test/ZNSAccessController.test.ts
```diff
@@ -279,4 +290,19 @@ describe("ZNSAccessController", () => {
...
+      );
+      expect(interfaceId).to.be.true;
+    });
+
+    it("should return false for unsupported interfaceId", async () => {
+      const unsupportedInterfaceId = "0xffffffff"; // Random unsupported interfaceId
+      const supports = await accessController.supportsInterface(unsupportedInterfaceId);
+      expect(supports).to.be.false;
+    });
+  });
```
Whytecrowe: redundant

test/ZNSAddressResolver.test.ts
```diff
@@ -204,6 +198,72 @@ describe("ZNSAddressResolver", () => {
...
+      ).to.emit(
+        zns.addressResolver,
+        "AccessControllerSet"
+      ).withArgs(zns.accessController.target);
+    });
+
+    it("should revert when a non-ADMIN tries to set AccessController", async () => {
+      await expect(
+        zns.addressResolver.connect(user).setAccessController(zns.accessController.target)
+      ).to.be.reverted;
```
Whytecrowe: check specific error

test/ZNSAddressResolver.test.ts
```diff
@@ -204,6 +198,72 @@ describe("ZNSAddressResolver", () => {
...
+    });
+
+    it("should revert when setting an AccessController as EOA address", async () => {
+      await expect(
+        zns.addressResolver.connect(deployer).setAccessController(user.address)
+      ).to.be.reverted;
```
Whytecrowe: specific error!

test/ZNSAddressResolver.test.ts
```diff
@@ -204,6 +198,72 @@ describe("ZNSAddressResolver", () => {
...
+    });
+
+    it("should revert when setting an AccessController as another non-AC contract address", async () => {
+      await expect(
+        zns.addressResolver.connect(deployer).setAccessController(zns.addressResolver.target)
+      ).to.be.reverted;
```
Whytecrowe: specific error

test/ZNSAddressResolver.test.ts
```diff
@@ -204,6 +198,72 @@ describe("ZNSAddressResolver", () => {
...
+    });
+
+    it("should revert when setting a zero address as AccessController", async () => {
+      await expect(
+        zns.addressResolver.connect(deployer).setAccessController(ethers.ZeroAddress)
+      ).to.be.reverted;
```
Whytecrowe: same here

test/ZNSCurvePricer.test.ts
```diff
@@ -872,30 +872,68 @@ describe("ZNSCurvePricer", () => {
...
-        ZERO_ADDRESS_ERR
-      );
+        "AccessControllerSet"
+      ).withArgs(zns.accessController.target);
+    });
+
+    it("should revert when a non-ADMIN tries to set AccessController", async () => {
+      await expect(
+        zns.curvePricer.connect(user).setAccessController(zns.accessController.target)
+      ).to.be.reverted;
```
Whytecrowe: search for all occurrences

PR #135, opened 2025-04-25
[v1.5] Polymorphic Price Config [WIP]

JamesEarle:
Change the price contracts to be stateless and use given bytes objects instead when calculating prices and fees. This allows domain registration and setup for selling subdomains to happen in a single transaction. Updated to include subregistrar changes in upgrade mock contracts


contracts/price/IZNSCurvePricer.sol
```diff
@@ -1,115 +1,77 @@
...
-
-    function setMaxLength(bytes32 domainHash, uint256 length) external;
-
-    function setCurveMultiplier(bytes32 domainHash, uint256 curveMultiplier) external;
+    error DivisionByZero();
 
-    function setPrecisionMultiplier(bytes32 domainHash, uint256 multiplier) external;
+    function encodeConfig(
+        CurvePriceConfig calldata config
+    ) external returns(bytes memory);
```
Whytecrowe: space after `returns` here and below

contracts/price/ZNSCurvePricer.sol
```diff
@@ -35,35 +32,74 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+        CurvePriceConfig memory config = CurvePriceConfig(
+            maxPrice,
+            curveMultiplier,
+            maxLength,
+            baseLength,
+            precisionMultiplier,
+            feePercentage
+        );
+
+        return config;
```
Whytecrowe: I think it could be:
```solidity
return CurvePriceConfig(
  maxPrice,
  curveMultiplier,
  maxLength,
  baseLength,
  precisionMultiplier,
  feePercentage
);
```

No need for the extra var

Whytecrowe: Also, I would fill as the full object with the props, just to be safe, even though it may look goofy

contracts/price/ZNSCurvePricer.sol
```diff
@@ -35,35 +32,74 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+     * @notice Validate the inputs for each variable in a price config
+     * @dev Will revert if incoming config is invalid
+     * 
+     * @param priceConfig The price config to evaluate
+     */
+    function validatePriceConfig(
+        bytes memory priceConfig
+    ) public override pure {
+        CurvePriceConfig memory config = decodePriceConfig(priceConfig);
+        _validatePriceConfig(config);
```
Whytecrowe: we could avoid the var here too if just pass decodePriceConfig(priceConfig) as a param to _validate...

contracts/price/ZNSCurvePricer.sol
```diff
@@ -75,280 +111,93 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
         bool skipValidityCheck
-    ) public view override returns (uint256) {
-        if (!priceConfigs[parentHash].isSet) revert ParentPriceConfigNotSet(parentHash);
+    ) public pure override returns (uint256) {
+        CurvePriceConfig memory config = decodePriceConfig(parentPriceConfig);
 
         if (!skipValidityCheck) {
             // Confirms string values are only [a-z0-9-]
             label.validate();
+            _validatePriceConfig(config);
```
Whytecrowe: this seems like an overkill. we shouldn't validate the config every time we need to get the price. we should validate it once before setting it in state and that's it.
Here we are going to pay for this every time a domain is registered, which there is no need for at all, and in general validating once before set in state is a much better design

contracts/price/ZNSCurvePricer.sol
```diff
@@ -75,280 +111,93 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
      * @param price The price to get the fee for
     */
     function getFeeForPrice(
-        bytes32 parentHash,
+        bytes memory parentPriceConfig,
         uint256 price
-    ) public view override returns (uint256) {
-        return (price * priceConfigs[parentHash].feePercentage) / PERCENTAGE_BASIS;
+    ) public pure override returns (uint256) {
+        CurvePriceConfig memory config = decodePriceConfig(parentPriceConfig);
```
Whytecrowe: could get rid of this var as well

contracts/price/ZNSCurvePricer.sol
Whytecrowe: remember, even if the function is pure, we still pay for all the gas it uses if it's called inside of a tx

contracts/price/ZNSCurvePricer.sol
Whytecrowe: by doing this you could also potentially make params to be stored in calldata

contracts/price/ZNSCurvePricer.sol
```diff
@@ -75,280 +111,93 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-            revert MaxLengthSmallerThanBaseLength(domainHash);
-
-        if (baseLength == 0 && config.curveMultiplier == 0)
-            revert DivisionByZero(domainHash);
-
-        priceConfigs[domainHash].baseLength = baseLength;
-    }
+        if (!skipValidityCheck) {
+            label.validate();
+            _validatePriceConfig(config);
```
Whytecrowe: same here. this is the wrong place for config validation

contracts/price/ZNSCurvePricer.sol
```diff
@@ -372,18 +221,18 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+        CurvePriceConfig memory config,
         uint256 length
-    ) internal view returns (uint256) {
-        CurvePriceConfig memory config = priceConfigs[parentHash];
-
+    ) internal pure returns (uint256) {
         // We use `maxPrice` as 0 to indicate free domains
         if (config.maxPrice == 0) return 0;
 
+        if (length == 0) return 0;
```
Whytecrowe, line 239: don't we do this in the top level function already?

contracts/price/ZNSFixedPricer.sol
```diff
@@ -1,44 +1,51 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.26;
 
-import { AAccessControlled } from "../access/AAccessControlled.sol";
-import { ARegistryWired } from "../registry/ARegistryWired.sol";
 import { IZNSFixedPricer } from "./IZNSFixedPricer.sol";
-import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import { StringUtils } from "../utils/StringUtils.sol";
 
+import { console } from "hardhat/console.sol";
```
Whytecrowe: remove this

contracts/price/ZNSFixedPricer.sol
```diff
@@ -1,44 +1,51 @@
...
-     * @param domainHash The hash of the domain who sets the price for subdomains
-     * @param _price The new price value set
-    */
-    function setPrice(bytes32 domainHash, uint256 _price) public override onlyOwnerOrOperator(domainHash) {
-        _setPrice(domainHash, _price);
+    function decodePriceConfig(
+        bytes memory priceConfig
+    ) public pure returns(PriceConfig memory) {
+        // console.log("in fp decodePriceConfig");
+        // console.logBytes(priceConfig);
```
Whytecrowe: remove this

contracts/price/ZNSFixedPricer.sol
```diff
@@ -50,126 +57,76 @@ contract ZNSFixedPricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-     * @dev `feePercentage` is set as a part of the `PERCENTAGE_BASIS` of 10,000 where 1% = 100
-     * @param domainHash The hash of the domain who sets the feePercentage for subdomains
-     * @param feePercentage The new feePercentage value set
-    */
-    function setFeePercentage(
-        bytes32 domainHash,
-        uint256 feePercentage
-    ) public override onlyOwnerOrOperator(domainHash) {
-        _setFeePercentage(domainHash, feePercentage);
+        return config.price;
```
Whytecrowe: same here. can we get rid of the `config` var?

contracts/price/ZNSFixedPricer.sol
```diff
@@ -50,126 +57,76 @@ contract ZNSFixedPricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-        bytes32 parentHash,
+        bytes memory parentPriceConfig,
         uint256 price
-    ) public view override returns (uint256) {
-        return (price * priceConfigs[parentHash].feePercentage) / PERCENTAGE_BASIS;
+    ) public pure override returns (uint256) {
+        // todo dont need 2 params for fixed price here, always same value
+        // but breaks interface in IZNSPricer otherwise
+        PriceConfig memory config = decodePriceConfig(parentPriceConfig);
+        return _getFeeForPrice(config, price);
```
Whytecrowe: same here.

contracts/registrar/IDistributionConfig.sol
```diff
@@ -36,9 +35,37 @@ interface IDistributionConfig {
...
+     * 
+     * @param pricerContract The address of the contract used for pricing subdomains
+     * @param paymentType The type of payment system used for selling subdomains 
+     * @param accessType The type of access that users have 
+     */
     struct DistributionConfig {
         IZNSPricer pricerContract;
         PaymentType paymentType;
         AccessType accessType;
+        bytes priceConfig;
```
Whytecrowe, line 49: would look better if both price vars are next to each other, but also look into packing here, we may be able to save on storage space

contracts/registrar/IDistributionConfig.sol
```diff
@@ -36,9 +35,37 @@ interface IDistributionConfig {
...
+    event AccessTypeSet(bytes32 indexed domainHash, AccessType accessType);
+
+    /**
+     * @notice Emitted when a new full `DistributionConfig` is set for a domain at once.
+     */
+    event DistributionConfigSet(
+        bytes32 indexed domainHash,
+        IZNSPricer pricerContract,
+        PaymentType paymentType,
+        AccessType accessType
```
Whytecrowe, line 70: missing priceConfig here

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -8,11 +8,17 @@ import { IZNSTreasury, PaymentConfig } from "../treasury/IZNSTreasury.sol";
...
 import { StringUtils } from "../utils/StringUtils.sol";
-import { ZeroAddressPassed, DomainAlreadyExists } from "../utils/CommonErrors.sol";
+import {
+    ZeroAddressPassed,
+    ZeroValuePassed,
+    AddressIsNotAContract,
+    DomainAlreadyExists
+} from "../utils/CommonErrors.sol";
 
+import { console } from "hardhat/console.sol";
```
Whytecrowe: remove

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -52,23 +59,36 @@ contract ZNSRootRegistrar is
...
         setTreasury(treasury_);
         setDomainToken(domainToken_);
     }
 
+    /**
+     * @notice Get the price of a root domain
+     * @param name The name for the domain
+     */
+    // TODO have this for users to more simply get names without providing encoded config
+    // but probably not necessary right? can just get config from here and call directly on pricer
```
Whytecrowe: remove comment

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -163,35 +182,54 @@ contract ZNSRootRegistrar is
     function _coreRegister(
         CoreRegisterArgs memory args
     ) internal {
+        // console.log("args.price: ", args.price);
```
Whytecrowe: remove

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -163,35 +182,54 @@ contract ZNSRootRegistrar is
...
         // payment part of the logic
         if (args.price > 0) {
             _processPayment(args);
         }
 
+        // console.log("after processPayment");
```
Whytecrowe: remove

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -163,35 +182,54 @@ contract ZNSRootRegistrar is
...
+
+
         // Get tokenId for the new token to be minted for the new domain
         uint256 tokenId = uint256(args.domainHash);
         // mint token
         domainToken.register(args.registrant, tokenId, args.tokenURI);
+        // console.log("after domainToken.register");
```
Whytecrowe: remove

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -163,35 +182,54 @@ contract ZNSRootRegistrar is
...
+                args.registrant,
+                ""
+            );
         }
 
         // Because we check in the web app for the existance of both values in a payment config,
         // it's fine to just check for one here
         if (args.paymentConfig.beneficiary != address(0)) {
+            // should be failing in this call
+            // console.log("before setPaymentConfig");
```
Whytecrowe: clean up

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -343,13 +381,27 @@ contract ZNSRootRegistrar is
...
-    function setRootPricer(address rootPricer_) public override onlyAdmin {
+    function setRootPricer(
+        address rootPricer_,
+        bytes memory rootPriceConfig_,
+        bool skipValidityCheck
+    ) public override onlyAdmin {
         if (rootPricer_ == address(0))
             revert ZeroAddressPassed();
 
+        if (rootPriceConfig_.length == 0) revert ZeroValuePassed();
```
Whytecrowe: can we not set zero price by giving empty config or does it need to be more specific?

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -343,13 +381,27 @@ contract ZNSRootRegistrar is
...
+
+        if (rootPricer_.code.length == 0) revert AddressIsNotAContract();
+
         rootPricer = IZNSPricer(rootPricer_);
 
-        emit RootPricerSet(rootPricer_);
+        if (!skipValidityCheck) {
```
Whytecrowe: idk why this is needed. imo it should always be validated when setting

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -343,13 +381,27 @@ contract ZNSRootRegistrar is
      * Only ADMIN in `ZNSAccessController` can call this function.
      * @param rootPricer_ Address of the IZNSPricer type contract to set as pricer of Root Domains
     */
-    function setRootPricer(address rootPricer_) public override onlyAdmin {
+    function setRootPricer(
```
Whytecrowe: if we do both, we should name the function better.

also, can we not just set the config separately?

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -68,6 +73,10 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
         setRootRegistrar(_rootRegistrar);
     }
 
+    function getPrice(bytes32 parentHash, string calldata label) external returns(uint256) {
+        //TODO
```
Whytecrowe: ???

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -31,7 +36,7 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
      * These configs are used to determine how subdomains are distributed for every parent.
      * @dev Note that the rules outlined in the DistributionConfig are only applied to direct children!
     */
-    mapping(bytes32 domainHash => DistributionConfig config) public override distrConfigs;
+    mapping(bytes32 domainHash => DistributionConfig config) public distrConfigs;
```
Whytecrowe: why remove the interface function?

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -98,10 +107,14 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
         if (registry.exists(domainHash))
             revert DomainAlreadyExists(domainHash);
 
+        // If this exists we know it has already been validated here
         DistributionConfig memory parentConfig = distrConfigs[parentHash];
 
+        if (address(parentConfig.pricerContract) == address(0))
+            revert ParentNotSet(parentHash);
```
Whytecrowe: confusing error name.

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -98,10 +107,14 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
+
         bool isOwnerOrOperator = registry.isOwnerOrOperator(parentHash, msg.sender);
-        if (parentConfig.accessType == AccessType.LOCKED && !isOwnerOrOperator)
+        if ((parentConfig.accessType == AccessType.LOCKED && !isOwnerOrOperator))
```
Whytecrowe: why double parentheses?

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -186,8 +199,15 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
 
+        if (config.priceConfig.length == 0)
+            revert ZeroValuePassed();
+
+        // Will revert if invalid
+        IZNSPricer(config.pricerContract).validatePriceConfig(config.priceConfig);
+
         distrConfigs[domainHash] = config;
 
+        // emit price config?
```
Whytecrowe: yes!

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -186,8 +199,15 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
         if (address(config.pricerContract) == address(0))
             revert ZeroAddressPassed();
 
+        if (config.priceConfig.length == 0)
+            revert ZeroValuePassed();
+
+        // Will revert if invalid
+        IZNSPricer(config.pricerContract).validatePriceConfig(config.priceConfig);
```
Whytecrowe, line 247: this is the correct place to validate!

contracts/registry/IZNSRegistry.sol
```diff
@@ -21,6 +21,8 @@ interface IZNSRegistry {
     struct DomainRecord {
         address owner;
         address resolver;
+        // address pricer;
+        // bytes priceConfig;
```
Whytecrowe: clean up

contracts/registry/IZNSRegistry.sol
```diff
@@ -43,6 +45,18 @@ interface IZNSRegistry {
...
+    //     bytes32 indexed domainHash,
+    //     address indexed pricer
+    // );
+
+    // // TODO natspec
+    // event DomainPriceConfigSet(
+    //     bytes32 indexed domainHash,
+    //     bytes indexed priceConfig
+    // );
+
```
Whytecrowe: ???

contracts/registry/IZNSRegistry.sol
```diff
@@ -115,9 +129,19 @@ interface IZNSRegistry {
         bytes32 domainHash
     ) external view returns (address);
 
+    // function getDomainPricer(
+    //     bytes32 domainHash
+    // ) external view returns (address);
+
+    // function getDomainPriceConfig(
+    //     bytes32 domainHash
+    // ) external view returns (bytes memory);
```
Whytecrowe: ???

contracts/registry/IZNSRegistry.sol
```diff
@@ -115,9 +129,19 @@ interface IZNSRegistry {
...
+
     function createDomainRecord(
         bytes32 domainHash,
         address owner,
+        // address pricer,
+        // bytes calldata priceConfig,
```
Whytecrowe: a lot of leftover comments

contracts/registry/IZNSRegistry.sol
```diff
@@ -137,6 +161,8 @@ interface IZNSRegistry {
     function updateDomainRecord(
         bytes32 domainHash,
         address owner,
+        // address pricer,
+        // bytes memory priceConfig,
```
Whytecrowe: clean up

contracts/registry/IZNSRegistry.sol
```diff
@@ -147,5 +173,11 @@ interface IZNSRegistry {
         string calldata resolverType
     ) external;
 
+    // function updateDomainPricerAndConfig(
+    //     bytes32 domainHash,
+    //     address pricer,
+    //     bytes memory priceConfig
+    // ) external;
```
Whytecrowe: ???

contracts/registry/ZNSRegistry.sol
```diff
@@ -146,6 +146,18 @@ contract ZNSRegistry is AAccessControlled, UUPSUpgradeable, IZNSRegistry {
...
+    //     bytes32 domainHash
+    // ) external view override returns (address) {
+    //     return records[domainHash].pricer;
+    // }
+
+    // function getDomainPriceConfig(
+    //     bytes32 domainHash
+    // ) external view override returns (bytes memory) {
+    //     return records[domainHash].priceConfig;
+    // }
```
Whytecrowe: more

contracts/registry/ZNSRegistry.sol
```diff
@@ -213,6 +225,8 @@ contract ZNSRegistry is AAccessControlled, UUPSUpgradeable, IZNSRegistry {
     function updateDomainRecord(
         bytes32 domainHash,
         address owner,
+        // address pricer,
+        // bytes memory priceConfig,
```
Whytecrowe: more

contracts/registry/ZNSRegistry.sol
```diff
@@ -254,6 +268,17 @@ contract ZNSRegistry is AAccessControlled, UUPSUpgradeable, IZNSRegistry {
...
+    // TODO natspec
+    // when changing pricer, must also change price config
+    // function updateDomainPricerAndConfig(
+    //     bytes32 domainHash,
+    //     address pricer,
+    //     bytes memory priceConfig
+    // ) external override onlyOwnerOrOperator(domainHash) {
+    //     _setDomainPricer(domainHash, pricer);
+    //     _setDomainPriceConfig(domainHash, priceConfig);
+    // }
```
Whytecrowe: more

contracts/registry/ZNSRegistry.sol
```diff
@@ -304,6 +329,28 @@ contract ZNSRegistry is AAccessControlled, UUPSUpgradeable, IZNSRegistry {
...
+
+    // function _setDomainPriceConfig(
+    //     bytes32 domainHash,
+    //     bytes memory priceConfig
+    // ) internal {
+    //     if (priceConfig.length == 0) revert ZeroValuePassed();
+
+    //     records[domainHash].priceConfig = priceConfig;
+    //     emit DomainPriceConfigSet(domainHash, priceConfig);
+    // }
```
Whytecrowe: mark these temp comment-outs with a `TODO <branch-name>: ...` so they are easy to catch when finishing a PR. this is what helps me not leave this stuff behind

contracts/upgrade-test-mocks/distribution/ZNSSubRegistrarMock.sol
```diff
@@ -132,14 +133,14 @@ contract ZNSSubRegistrarUpgradeMock is
...
                     true
                 );
             } else {
                 coreRegisterArgs.price = IZNSPricer(address(parentConfig.pricerContract))
                     .getPrice(
-                    parentHash,
+                    parentConfig.priceConfig,
                     label,
                     true
                 );
```
Whytecrowe, line 151: need to copy the rest of the new code from SubRegistrar here

src/deploy/missions/contracts/curve-pricer.ts
```diff
@@ -16,22 +16,10 @@ IZNSCampaignConfig,
 IZNSContracts
 > {
   proxyData = {
-    isProxy: true,
+    isProxy: false,
     kind: ProxyKinds.uups,
```
Whytecrowe: remove this line

src/deploy/missions/contracts/root-registrar.ts
```diff
@@ -37,6 +38,7 @@ IZNSContracts
       await registry.getAddress(),
       // we use CurvePricer as the IZNSPricer for root domains
       await curvePricer.getAddress(),
+      DEFAULT_CURVE_PRICE_CONFIG_BYTES,
```
Whytecrowe: this is incorrect! this should come from the getConfig function that reads env vars and then encodes whatever we pass in them at deploy time then passes here to the contract.

Whytecrowe: also add test defaults as ENV vars to the proper default config where all the default test values live.

test/MochaTesting.test.ts
```diff
@@ -0,0 +1,551 @@
+import * as hre from "hardhat";
+import { assert, expect } from "chai";
+import { ethers } from "ethers";
+import { SignerWithAddress } from "@nomicfoundation/hardhat-ethers/signers";
+import { AccessType, DEFAULT_CURVE_PRICE_CONFIG_BYTES, DEFAULT_FIXED_PRICER_CONFIG_BYTES, DEFAULT_TOKEN_URI, deployZNS, distrConfigEmpty, DOMAIN_EXISTS_ERR, encodePriceConfig, IDistributionConfig, IDomainConfigForTest, IFixedPriceConfig, IPathRegResult, IZNSContractsLocal, NOT_OWNER_OF_ERR, paymentConfigEmpty, PaymentType } from "./helpers";
+import { registrationWithSetup } from "./helpers/register-setup";
+import { registerDomainPath, validatePathRegistration } from "./helpers/flows/registration";
+
+// todo is state change only related to hardhat? or mocha generally?
```
Whytecrowe: ?

test/MochaTesting.test.ts
```diff
@@ -0,0 +1,551 @@
+import * as hre from "hardhat";
```
Whytecrowe: not sure what this file is supposed to be. also, check file name

test/ZNSCurvePricer.test.ts
```diff
@@ -42,6 +46,8 @@ describe("ZNSCurvePricer", () => {
   let zns : IZNSContractsLocal;
   let domainHash : string;
 
+  let utils : Utils; // ?
```
Whytecrowe: ?

test/ZNSCurvePricer.test.ts
```diff
@@ -237,781 +231,161 @@ describe("ZNSCurvePricer", () => {
...
-    });
-
-    it("Correctly sets max price", async () => {
-      const newMaxPrice = DEFAULT_PRICE_CONFIG.maxPrice + ethers.parseEther("553");
-      await zns.curvePricer.connect(user).setMaxPrice(domainHash, newMaxPrice);
 
-      const params = await zns.curvePricer.priceConfigs(domainHash);
-      expect(params.maxPrice).to.eq(newMaxPrice);
+      // TODO seems like its somehow failing outside regular test framework?
+      // "VM exception while processing transaction:..."
```
Whytecrowe: ?

test/ZNSCurvePricer.test.ts
```diff
@@ -237,781 +231,161 @@ describe("ZNSCurvePricer", () => {
...
-
-      const paramsAfter = await zns.curvePricer.priceConfigs(domainHash);
-
-      const expectedPriceAfter = await getCurvePrice(defaultDomain, newConfig2);
-      const priceAfter = await zns.curvePricer.getPrice(domainHash, defaultDomain, true);
-      expect(priceAfter).to.eq(expectedPriceAfter);
-      expect(priceAfter).to.not.eq(paramsAfter.maxPrice);
-    });
-
-    it("Returns the price = 0 whenever the baseLength is 0", async () => {
```
Whytecrowe, line 739: please review some of these tests. some of them still seem valid

test/ZNSRootRegistrar.test.ts
```diff
@@ -84,7 +92,8 @@ describe("ZNSRootRegistrar", () => {
 
     zns = campaign.state.contracts;
 
-    // await zns.accessController.connect(deployer).grantRole(DOMAIN_TOKEN_ROLE, await zns.domainToken.getAddress());
+    // TODO put this back
```
Whytecrowe: ?

test/ZNSRootRegistrar.test.ts
```diff
@@ -601,11 +615,13 @@ describe("ZNSRootRegistrar", () => {
     });
 
     it("Successfully registers a domain with distrConfig and adds it to state properly", async () => {
-      const distrConfig = {
+      const distrConfig : IDistributionConfig = {
         pricerContract: await zns.fixedPricer.getAddress(),
+        priceConfig: ZeroHash, // TODO fix by adding helper
```
Whytecrowe: ?

test/ZNSSubRegistrar.test.ts
```diff
@@ -102,22 +124,32 @@ describe("ZNSSubRegistrar", () => {
         feePercentage: BigInt(0),
       };
 
+      // TODO testing this
```
Whytecrowe: ?

test/ZNSSubRegistrar.test.ts
```diff
@@ -173,22 +205,34 @@ describe("ZNSSubRegistrar", () => {
...
+        //   lvl2SubOwner.address,
+        //   subTokenURI,
+        //   distrConfigEmpty,
+        //   {
+        //     token: await zns.meowToken.getAddress(),
+        //     beneficiary: rootOwner.address,
+        //   });
+      // } catch (e) {
+      //   console.log((e as Error).message);
+      // }
```
Whytecrowe: ???

test/ZNSSubRegistrar.test.ts
```diff
@@ -1055,22 +1099,38 @@ describe("ZNSSubRegistrar", () => {
     });
 
     it("should register a new 2 lvl path at lvl 3 of the existing path", async () => {
-      const newConfigs = [
+      // TODO FINDME func for default with only needed values changes to simplify
```
Whytecrowe: ???

test/ZNSSubRegistrar.test.ts
```diff
@@ -1143,58 +1203,62 @@ describe("ZNSSubRegistrar", () => {
...
-      ).to.be.revertedWithCustomError(
-        zns.registry,
-        NOT_AUTHORIZED_ERR
-      );
+      // await expect(
+      //   zns.registry.connect(lvl3SubOwner).updateDomainRecord(lvl3Hash, rootOwner.address, lvl4SubOwner.address)
+      // ).to.be.revertedWithCustomError(
+      //   zns.registry,
+      //   NOT_AUTHORIZED_ERR
+      // );
```
Whytecrowe: ??? - all of the commented lines above

test/ZNSSubRegistrar.test.ts
```diff
@@ -1339,16 +1403,19 @@ describe("ZNSSubRegistrar", () => {
...
             accessType: AccessType.MINTLIST,
           },
           paymentConfig: {
             token: await zns.meowToken.getAddress(),
             beneficiary: branchLvl1Owner.address,
           },
-          priceConfig: { price: fixedPrice, feePercentage: BigInt(0) },
         },
-      });
+      } as IRegisterWithSetupArgs);
```
Whytecrowe: why is this needed?

test/ZNSSubRegistrar.test.ts
```diff
@@ -2924,16 +2995,25 @@ describe("ZNSSubRegistrar", () => {
...
-    let fixedFeePercentage : bigint;
+    let randomAcc : SignerWithAddress;
+
+    let fixedPrice : bigint = ethers.parseEther("397");
+    let fixedFeePercentage : bigint = BigInt(200);
+
+    let fixedPriceConfig : IFixedPriceConfig = {
+      price: fixedPrice,
+      feePercentage: fixedFeePercentage
+    }
```
Whytecrowe: set eslint --fix on save to avoid leaving lines like this with no ;

test/ZNSSubRegistrar.test.ts
```diff
@@ -2948,19 +3028,16 @@ describe("ZNSSubRegistrar", () => {
         lvl4SubOwner,
         lvl5SubOwner,
         lvl6SubOwner,
+        randomAcc
```
Whytecrowe: same here with eslint. should be a trailing comma here

test/ZNSSubRegistrar.test.ts
```diff
@@ -3032,97 +3109,108 @@ describe("ZNSSubRegistrar", () => {
...
-          DEFAULT_TOKEN_URI,
-          domainConfigs[1].fullConfig.distrConfig,
-          paymentConfigEmpty
-        )
-      ).to.be.revertedWithCustomError(
-        zns.subRegistrar,
-        DOMAIN_EXISTS_ERR
-      );
+      const dOwner = await zns.registry.getDomainOwner(regResults[1].domainHash);
+      console.log(`dhash[1] earlier: ${regResults[1].domainHash}`);
```
Whytecrowe: clean up

test/ZNSSubRegistrar.test.ts
```diff
@@ -3032,97 +3109,108 @@ describe("ZNSSubRegistrar", () => {
...
-      await zns.domainToken.connect(lvl2SubOwner).updateTokenOwner(
-        lvl2SubOwner.address,
-        lvl3SubOwner.address,
-        regResults[1].domainHash
-      );
-
-      // fail again
-      await expect(
-        zns.rootRegistrar.connect(lvl2SubOwner).revokeDomain(
+    // describe("another inner describe for testing", () => {
```
Whytecrowe: clean up

test/ZNSSubRegistrar.test.ts
```diff
@@ -3032,97 +3109,108 @@ describe("ZNSSubRegistrar", () => {
...
-      expect(dataFromRegAfterReclaim.resolver).to.eq(ethers.ZeroAddress);
-    });
+        const dataFromRegAfterReclaim = await zns.registry.getDomainRecord(regResults[1].domainHash);
+        expect(dataFromRegAfterReclaim.owner).to.eq(lvl2SubOwner.address);
+        expect(dataFromRegAfterReclaim.resolver).to.eq(ethers.ZeroAddress);
+      });
+    // });
 
-    describe("#setDistributionConfigForDomain()", () => {
+    // describe("#setDistributionConfigForDomain()", () => {
```
Whytecrowe: clean up

test/ZNSSubRegistrar.test.ts
```diff
@@ -3032,97 +3109,108 @@ describe("ZNSSubRegistrar", () => {
...
 
-      const dataFromRegAfterReclaim = await zns.registry.getDomainRecord(regResults[1].domainHash);
-      expect(dataFromRegAfterReclaim.owner).to.eq(lvl2SubOwner.address);
-      expect(dataFromRegAfterReclaim.resolver).to.eq(ethers.ZeroAddress);
-    });
+        const dataFromRegAfterReclaim = await zns.registry.getDomainRecord(regResults[1].domainHash);
+        expect(dataFromRegAfterReclaim.owner).to.eq(lvl2SubOwner.address);
+        expect(dataFromRegAfterReclaim.resolver).to.eq(ethers.ZeroAddress);
+      });
+    // });
```
Whytecrowe: clean up

test/ZNSSubRegistrar.test.ts
```diff
@@ -3032,97 +3109,108 @@ describe("ZNSSubRegistrar", () => {
...
 
-    describe("#setDistributionConfigForDomain()", () => {
+    // describe("#setDistributionConfigForDomain()", () => {
       it("should re-set distribution config for an existing subdomain", async () => {
+        //TODO FINDMEDEBUG3211
```
Whytecrowe: clean up

test/ZNSSubRegistrar.test.ts
```diff
@@ -3198,49 +3291,196 @@ describe("ZNSSubRegistrar", () => {
...
+    // describe("#setPricerContractForDomain()", () => {
+      // let domainHash : string;
+      // // let user : SignerWithAddress;
+
+      // before(async () => {
+      //   domainHash = regResults[1].domainHash;
+      //   // [ user ] = await hre.ethers.getSigners()
+      // })
+
+      // #setPricerContractForDomain()
```
Whytecrowe: clean up

test/ZNSSubRegistrar.test.ts
```diff
@@ -3198,49 +3291,196 @@ describe("ZNSSubRegistrar", () => {
...
+          feePercentage: DEFAULT_PROTOCOL_FEE_PERCENT,
+        };
+
+        const asBytes = encodePriceConfig(newConfig);
+
+        const domainHash = regResults[1].domainHash;
+        console.log(`dhash[1]: ${domainHash}`);
+
+        // as a user of "domainHash" that's not 0x0
+        // TODO FINDME NOT AUTHORIZED?
```
Whytecrowe: clean up

test/gas/gas-costs.json
```diff
@@ -1,4 +1,4 @@
 {
-	"Root Domain Price": "464759",
-	"Subdomain Price": "458275"
+	"Root Domain Price": "535756",
+	"Subdomain Price": "541376"
```
Whytecrowe: oh, wow! what's adding so much gas? validation + storage of the config?

test/helpers/flows/registration.ts
```diff
@@ -2,20 +2,31 @@
...
+}
 
 export const registerDomainPath = async ({
   zns,
   domainConfigs,
 } : {
   zns : IZNSContractsLocal;
   domainConfigs : Array<IDomainConfigForTest>;
-}) => domainConfigs.reduce(
+}) : Promise<Array<IPathRegResult>> => domainConfigs.reduce( // TODO how to provide return type?
```
Whytecrowe: clean up

test/helpers/pricing.ts
```diff
@@ -84,4 +86,118 @@ export const getPriceObject = (
     expectedPrice,
     stakeFee,
   };
-};
\ No newline at end of file
+};
+
+// TODO can we make one single fucntion to route for both?
```
Whytecrowe: probably. could add an enum that has CURVE, FIXED, and can be extended later

test/helpers/pricing.ts
```diff
@@ -84,4 +86,118 @@ export const getPriceObject = (
...
+
+export const encodePriceConfig = (
+  config : ICurvePriceConfig | IFixedPriceConfig,
+) => {
+  if (Object.keys(config).length > 2) {
+    return encodeCurvePriceConfig(config as ICurvePriceConfig);
+  } else {
+    return encodeFixedPriceConfig(config as IFixedPriceConfig);
+  }
+}
```
Whytecrowe: or like that

test/helpers/pricing.ts
```diff
@@ -84,4 +86,118 @@ export const getPriceObject = (
...
+
+export const decodePriceConfig = (
+  config : string
+) => {
+  if (config.length > DEFAULT_FIXED_PRICER_CONFIG_BYTES.length) {
+    return decodeCurvePriceConfig(config);
+  } else {
+    return decodeFixedPriceConfig(config);
+  }
+}
```
Whytecrowe: seems like this needs a linter fix pass

test/helpers/register-setup.ts
```diff
@@ -101,6 +107,8 @@ export const defaultSubdomainRegistration = async ({
 }) => {
   const supplyBefore = await zns.domainToken.totalSupply();
 
+
+  // console.log(distrConfig);
```
Whytecrowe: clean up

test/helpers/register-setup.ts
```diff
@@ -179,21 +178,22 @@ export const registrationWithSetup = async ({
...
-      },
-    );
+    // TODO uncomment and fix these!
+    // await zns.fixedPricer.connect(user).setPriceConfig(
+    //   domainHash,
+    //   {
+    //     ...fullConfig.priceConfig as IFixedPriceConfig,
+    //     isSet: true,
+    //   },
+    // );
```
Whytecrowe: clean up

test/helpers/register-setup.ts
```diff
@@ -179,21 +178,22 @@ export const registrationWithSetup = async ({
...
-        isSet: true,
-      },
-    );
+    // await zns.curvePricer.connect(user).setPriceConfig(
+    //   domainHash,
+    //   {
+    //     ...fullConfig.priceConfig as ICurvePriceConfig,
+    //     isSet: true,
+    //   },
+    // );
```
Whytecrowe: clean up

test/helpers/types.ts
```diff
@@ -97,38 +100,50 @@ export interface DeployZNSParams {
...
 export interface IDistributionConfig {
-  pricerContract : string;
+  pricerContract : string | Addressable;
+  priceConfig : string;
   paymentType : bigint;
   accessType : bigint;
 }
 
+export interface IDistributionConfigExtended extends Typed {
+  [k: string] : any;
```
Whytecrowe: what is the point for a type like this?

test/helpers/utils.ts
```diff
@@ -31,4 +34,169 @@ export const getRandomString = (length : number) => {
...
+//         domainLabels[i],
+//         parentHashes[i],
+//         distrConfigs[i],
+//         paymentConfigs[i],
+//       )
+//     );
+//   }
+
+//   return configs;
+// }
```
Whytecrowe: clean up

test/helpers/utils.ts
```diff
@@ -31,4 +34,169 @@ export const getRandomString = (length : number) => {
...
+  distrConfig ?: Partial<IDistributionConfig>,
+  paymentConfig ?: Partial<IPaymentConfig>
+) : Promise<IDomainConfigForTest> => {
+
+  let distrConfigToUse : IDistributionConfig;
+  let paymentConfigToUse : IPaymentConfig;
+
+  // TODO could technically break this by handing independent pieces that dont mesh
+  // like curve pricer address and price config for fixed pricer
+  // but this SHOULD fail when using it because of validation in pricer contracts
```
Whytecrowe: clean up

Whytecrowe (changes requested):
please, address comments

contracts/price/ZNSCurvePricer.sol
JamesEarle: yeah was unsure on this one too, it's been removed

contracts/price/ZNSCurvePricer.sol
JamesEarle: yep! This was leftover trying to test something

contracts/registrar/IDistributionConfig.sol
JamesEarle: You can't do packing between multiple structs, only for a struct itself. So if a higher level struct has multiple inner structs those don't share storage and can't be optimized for storage in combination with each other

contracts/registrar/ZNSRootRegistrar.sol
JamesEarle: This is to protect against nothing be passed, like if you just submitted an empty string. If you wanted zero price you could submit `ZeroHash` or just `0x0` to have a length higher than 0

contracts/registrar/ZNSRootRegistrar.sol
JamesEarle: I've added a function set just the config separately, but the setter for the pricer contract to use should also still set both. if changing the pricer, there will inherently be a new config

src/deploy/missions/contracts/curve-pricer.ts
JamesEarle: the object shows a type error if I try to leave it unspecified instead of just marking as false

test/ZNSSubRegistrar.test.ts
JamesEarle: Because adding typing makes it show errors when it changes in the future, so it is obvious what's failing ahead of time

src/deploy/missions/contracts/root-registrar.ts
```diff
@@ -7,6 +7,7 @@ import { znsNames } from "./names";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { SignerWithAddress } from "@nomicfoundation/hardhat-ethers/signers";
 import { IZNSCampaignConfig, IZNSContracts } from "../../campaign/types";
+import { DEFAULT_CURVE_PRICE_CONFIG_BYTES } from "../../../../test/helpers";
```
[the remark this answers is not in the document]
Whytecrowe: I agree with this. this should be in a different place in the env.ts file where all the default values are

src/deploy/missions/contracts/curve-pricer.ts
Whytecrowe: kind: ProxyKinds.uups - this line. it's not needed here. check AccessController mission - that contract is not a proxy

test/helpers/register-setup.ts
```diff
@@ -59,18 +57,23 @@ export const approveForParent = async ({
   domainLabel : string;
 }) => {
   const { pricerContract } = await zns.subRegistrar.distrConfigs(parentHash);
+  const { priceConfig } = await zns.subRegistrar.distrConfigs(parentHash);
```
Whytecrowe: why the whole separate line, just add `priceConfig` to the destructuring above

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -98,8 +102,12 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
         if (registry.exists(domainHash))
             revert DomainAlreadyExists(domainHash);
 
+        // If this exists we know it has already been validated here
         DistributionConfig memory parentConfig = distrConfigs[parentHash];
 
+        if (address(parentConfig.pricerContract) == address(0))
+            revert ParentNotSetupForDistribution(parentHash);
```
Whytecrowe: do we actually need this check? what if I register as owner of parent and my domain is locked, but I still want to make subs for myself. there should be a test for it.
I don't think this is a good line to add. am I wrong?

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -179,18 +187,30 @@ contract ZNSRootRegistrar is
         // If the `domainAddress` is not provided upon registration, a user can call `ZNSAddressResolver.setAddress`
         // to set the address themselves.
         if (args.domainAddress != address(0)) {
-            registry.createDomainRecord(args.domainHash, args.registrant, "address");
+
+            registry.createDomainRecord(
```
Whytecrowe: why empty line here above?

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -179,18 +187,30 @@ contract ZNSRootRegistrar is
...
+                args.domainHash, 
+                args.registrant,
+                ""
+            );
         }
 
         // Because we check in the web app for the existance of both values in a payment config,
         // it's fine to just check for one here
         if (args.paymentConfig.beneficiary != address(0)) {
+            // should be failing in this call
```
Whytecrowe: what is this about?

contracts/price/ZNSCurvePricer.sol
```diff
@@ -35,35 +32,76 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
-        _setAccessController(accessController_);
-        _setRegistry(registry_);
-
-        setPriceConfig(0x0, zeroPriceConfig_);
+    function validatePriceConfig(
+        bytes memory priceConfig
+    ) public override pure {
+        // No need to check incoming bytes length here like we do in `fixedPricer`
+        // If incorrect bytes length, downstream errors are thrown
+        _checkLength(priceConfig);
```
Whytecrowe, line 101: comment above says "No need to check incoming bytes length..." but this calls `_checkLength()`.
??

contracts/price/ZNSCurvePricer.sol
```diff
@@ -75,280 +113,101 @@ contract ZNSCurvePricer is AAccessControlled, ARegistryWired, UUPSUpgradeable, I
...
+        return (price * feePercentage) / PERCENTAGE_BASIS;
     }
 
-    function _validateSetPrecisionMultiplier(
-        bytes32 domainHash,
-        uint256 multiplier
-    ) internal {
-        if (multiplier == 0 || multiplier > 10**18) revert InvalidPrecisionMultiplierPassed(domainHash);
+    function _validatePriceConfig(CurvePriceConfig memory config) internal pure {
+        if (config.curveMultiplier == 0 && config.baseLength == 0)
```
Whytecrowe, line 192: seems like this check is missing here:
```solidity
if (curveMultiplier == 0 && config.baseLength == 0)
            revert DivisionByZero(domainHash);
```

contracts/price/ZNSCurvePricer.sol
JamesEarle: was out of date, removed it

contracts/price/ZNSCurvePricer.sol
JamesEarle: ? That "missing" check is the check you commented on?

contracts/registrar/ZNSRootRegistrar.sol
JamesEarle: was for debugging, removed

contracts/price/ZNSCurvePricer.sol
Whytecrowe: haha my bad! I was looking in the wrong place it seems :))

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -9,7 +9,7 @@ import { ARegistryWired } from "../registry/ARegistryWired.sol";
 import { StringUtils } from "../utils/StringUtils.sol";
 import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import {
-    DomainAlreadyExists,
+    DomainAlreadyExists, // TODO used?
```
Whytecrowe: left comment

test/ZNSSubRegistrar.test.ts
```diff
@@ -1625,7 +1635,7 @@ describe("ZNSSubRegistrar", () => {
           tokenURI: DEFAULT_TOKEN_URI,
           distrConfig: distrConfigEmpty,
           paymentConfig: paymentConfigEmpty,
-        })
+        } as ISubRegistrarConfig)
```
Whytecrowe: why do we need these? seems like a redundant addition that would only bring more work.
1. this type already exists in the typechain contract, no need to create a new one.
2. type should be inferred from the function itself and has no need to specify it manually in every call

test/ZNSSubRegistrar.test.ts
```diff
@@ -3693,6 +3704,14 @@ describe("ZNSSubRegistrar", () => {
...
         feePercentage: DEFAULT_PROTOCOL_FEE_PERCENT,
       };
+      // TODO come back to this
+      //   await expect(
+      //     zns.subRegistrar.connect(deployer).setDistributionConfigForDomain(domainHash, newConfig)
+      //   ).to.be.revertedWithCustomError(
+      //     zns.subRegistrar,
+      //     NOT_AUTHORIZED_ERR
+      //   );
+      // });
```
Whytecrowe: ?

test/helpers/types.ts
```diff
@@ -140,6 +142,18 @@ export interface IRegisterWithSetupArgs extends ConfigArgsBase {
...
+export interface DefaultRootRegistrationArgs {
+  user : SignerWithAddress;
+  zns : IZNSContractsLocal | IZNSContracts;
+  domainName : string;
+  tokenOwner ?: string;
+  domainContent ?: string;
+  tokenURI ?: string;
+  distrConfig ?: IDistributionConfig;
+  paymentConfig ?: IPaymentConfig;
+}
```
Whytecrowe, line 163: why do we need this type? the type for a struct already exists in typechain

Whytecrowe (changes requested):
comments

test/ZNSSubRegistrar.test.ts
JamesEarle: I have removed it here, but it doesn't seem typechain is generating this type.

src/tenderly/devnet/run-all-flows.ts
```diff
@@ -50,6 +51,7 @@ export const runAllFlows = async () => {
 
   const rootHash = await registrationWithSetup({
     zns,
+    tokenOwner: governor.address,
```
Whytecrowe: this really only needs to be passed if you want token owner to be different from the `user` and domain owner. otherwise can skip this prop

src/tenderly/devnet/run-all-flows.ts
```diff
@@ -75,10 +77,11 @@ export const runAllFlows = async () => {
   await registrationWithSetup({
     zns,
     user,
+    tokenOwner: user.address,
     parentHash: rootHash,
     domainLabel: subdomainLabel,
     fullConfig: fullSubConfig,
-  });
+  } as IRegisterWithSetupArgs);
```
Whytecrowe: this seems unnecessary, it should already match with the param

contracts/price/IZNSCurvePricer.sol
```diff
@@ -1,115 +1,71 @@
...
 
 
-interface IZNSCurvePricer is ICurvePriceConfig, IZNSPricer {
+interface IZNSCurvePricer is IZNSPricer {
 
     /**
-     * @notice Reverted when multiplier passed by the domain owner
-     * is equal to 0 or more than 10^18, which is too large.
+     * @notice Struct for each configurable variable for price calculations.
+     * Does NOT include variables for calcs of registration fees.
```
Whytecrowe: weird comment and may be incorrect

PR #136, opened 2025-04-29
[v1.5] SubdomainToken contract

MichaelKorchagin:


.gitignore
JamesEarle, line 1: If files show like this, where the entire file is changed, it is usually because of the line endings for the file changing and it makes it really hard to review what *actually* changed. Please always use LF not CRLF to be consistent with the repo

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+        address registry_,
+        address accessController_
+    )
+        ERC721(name_, symbol_)
+        EIP712(name_, version_)
+    {
+        _setRegistry(registry_);
+    }
+
+    function register(address to, uint256 tokenId, string memory _tokenURI)
```
JamesEarle: To remain consistent with the other functions in contracts we define in the repository, when line breaks are necessary put them within parameter lists instead.

```solidity
function myFunc(
  address var1,
  address var2,
  uint256 var3
) external {
  ...
}
```

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+        external
+    {
+        ++_totalSupply;
+        _safeMint(to, tokenId);
+        _setTokenURI(tokenId, _tokenURI);
+    }
+
+    function revoke(uint256 tokenId)
```
JamesEarle: for functions without long signatures like this, it isn't necessary to break the signature into multiple lines.

```solidity
function revoke(uint256 tokenID) external {
  ...
}
```

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    function setRegistry(address registry_)
+        external
+        override(ARegistryWired)
+    {
+        _setRegistry(registry_);
+    }
+
+    // TODO: add access control, when it's ready
+    // TODO: do we check the passed string?
```
JamesEarle: No we don't check this on-chain

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+    {
+        ++_totalSupply;
+        _safeMint(to, tokenId);
+        _setTokenURI(tokenId, _tokenURI);
+    }
+
+    function revoke(uint256 tokenId)
+        external
+    {
+        _burn(tokenId);
```
JamesEarle: Be sure access control is on this function as well, when it is ready to be implemented.

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+        uint256 tokenId
+    ) 
+        public 
+        override(ERC721, IERC721) 
+    {
+        // Transfer the token
+        super.transferFrom(from, to, tokenId);
+
+        // TODO: make a fucntion or update existing one in the registry
+        // registry.updateDomainOwner(bytes32(abi.encodePacked(tokenId)), to);
```
JamesEarle: Should this be uncommented?

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    // TODO: add access control, when it's ready
+    function setTokenURI(uint256 tokenId, string memory _tokenURI)
+        external
+    {
+        _setTokenURI(tokenId, _tokenURI);
+        emit TokenURISet(tokenId, _tokenURI);
+    }
+
+    function transferFrom(
```
JamesEarle, line 59: also be sure to include AC here when ready

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+    EIP712,
+    ERC721Votes,
+    ERC2981,
+    ARegistryWired {
+
+    uint256 private _totalSupply;
+
+    string private baseURI;
+
+    event TokenURISet(uint256 indexed tokenId, string tokenURI);
```
JamesEarle: Put events in the interface `IZNSSubdomainToken`

contracts/token/ZNSSubdomainToken.sol
JamesEarle, line 1: After addressing other things I've pointed out, add natspec comments

contracts/token/ZNSSubdomainToken.sol
```diff
@@ -0,0 +1,177 @@
...
+
+
+contract ZNSSubdomainToken is
+    ERC721,
+    ERC721Enumerable,
+    ERC721URIStorage,
+    EIP712,
+    ERC721Votes,
+    ERC2981,
+    ARegistryWired {
```
Whytecrowe: why doesn't this just inherit the ZNSDomainToken contract? It seems to have almost the same implementation. is there a reason you chose to make this separate?

contracts/token/IZNSSubdomainToken.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+interface ISubdomainToken {
```
Whytecrowe, line 4: always a 2-line break before contract declaration

PR #137, opened 2025-05-16
[v1.5] Controlled Domains

Whytecrowe:


contracts/registrar/IZNSRootRegistrar.sol
```diff
@@ -12,11 +12,12 @@ import { PaymentConfig } from "../treasury/IZNSTreasury.sol";
...
     bytes32 parentHash;
     bytes32 domainHash;
-    address registrant;
-    address domainAddress;
+    string label;
+    address domainOwner;
+    address tokenOwner;
     uint256 price;
     uint256 stakeFee;
-    string label;
```
JamesEarle, line 19: why do we move label? we get rid of any potential gain from struct memory packing by having it in between addresses

contracts/registrar/ZNSRootRegistrar.sol
```diff
@@ -288,45 +304,34 @@ contract ZNSRootRegistrar is
...
-        if (!isOwnerOf(domainHash, msg.sender, OwnerOf.TOKEN))
-            revert NotTheOwnerOf(OwnerOf.TOKEN, msg.sender, domainHash);
-
-        registry.updateDomainOwner(domainHash, msg.sender);
+        if (msg.sender != registry.getDomainOwner(domainHash))
+            revert NotAuthorizedForDomain(msg.sender, domainHash);
 
-        emit DomainReclaimed(domainHash, msg.sender);
-    }
+        address curTokenOwner = domainToken.ownerOf(uint256(domainHash));
```
JamesEarle, line 346: can optimize by not assigning to a variable

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -74,71 +73,77 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
-        bool isOwnerOrOperator = registry.isOwnerOrOperator(parentHash, msg.sender);
-        if (parentConfig.accessType == AccessType.LOCKED && !isOwnerOrOperator)
-            revert ParentLockedOrDoesntExist(parentHash);
-
-        if (parentConfig.accessType == AccessType.MINTLIST) {
+     */
+    function registerSubdomain(SubdomainRegisterArgs calldata regArgs) external override returns (bytes32) {
+        address domainRecordOwner = msg.sender;
+        address parentOwner = registry.getDomainOwner(regArgs.parentHash);
+        bool isOwner = msg.sender == parentOwner;
```
JamesEarle, line 98: small gas savings here as well by not assigning this to variable

contracts/registrar/ZNSSubRegistrar.sol
```diff
@@ -74,71 +73,77 @@ contract ZNSSubRegistrar is AAccessControlled, ARegistryWired, UUPSUpgradeable,
...
-
-        DistributionConfig memory parentConfig = distrConfigs[parentHash];
-
-        bool isOwnerOrOperator = registry.isOwnerOrOperator(parentHash, msg.sender);
-        if (parentConfig.accessType == AccessType.LOCKED && !isOwnerOrOperator)
-            revert ParentLockedOrDoesntExist(parentHash);
-
-        if (parentConfig.accessType == AccessType.MINTLIST) {
+     */
+    function registerSubdomain(SubdomainRegisterArgs calldata regArgs) external override returns (bytes32) {
```
JamesEarle: I think `args` would suffice as a variable name and be slightly simpler

contracts/registrar/IZNSRootRegistrar.sol
Whytecrowe: There's no packing here since this is a memory only struct

contracts/registrar/ZNSRootRegistrar.sol
Whytecrowe: var used twice here. memory var is cheaper than another external call

contracts/registrar/ZNSSubRegistrar.sol
Whytecrowe: var used twice here. memory var is cheaper than another external call

PR #138, opened 2025-05-28
[v1.5] Limited Access to Registrations

Whytecrowe:
This PR needs to be merged with the base branch AFTER all the previous PRs have been merged (Bulk Registration, Controlled Domains specifically).

After that:

1. Add access modifiers to the bulk registration functions
2. Fix conflicts with turning domain registration params
3. Make params to registerRootDomain() a struct as well
4. Unblock the tests and make sure everything works without extra additions (tests for struct params should be fixed in Controlled Domains PR)

DONE


contracts/registrar/ARegistrationPause.sol
```diff
@@ -0,0 +1,52 @@
...
+     * @notice Boolean flag to pause public registration of new domains.
+     * @dev When this flag is active, only ZNS ADMINs can register new domains.
+     */
+    bool public registrationPaused;
+
+    /**
+     * @notice Modifier to make a function publicly callable only when registration is not paused.
+     * If registration is paused, only ADMINs can call the function.
+     */
+    modifier whenRegNotPaused(IZNSAccessController accessController) {
```
JamesEarle, line 34: I think `whenNotPaused` is sufficient as a name, we don't need `reg` included, that context is provided by the contract and file name already

contracts/registrar/ARegistrationPause.sol
```diff
@@ -0,0 +1,52 @@
...
+        if (registrationPaused) {
+            if (!accessController.isAdmin(msg.sender))
+                revert PublicRegistrationPaused();
+        }
+        _;
+    }
+
+    function _setRegistrationPause(
```
JamesEarle, line 42: similarly I think `Registration` may not be needed here, it is apparent what it is pausing already

test/helpers/register-setup.ts
JamesEarle, line 1: it will help speed things up as we refactor tests to also focus on using `zns.[contractName].target` instead of the await calls to `getAddress()`

test/helpers/register-setup.ts
```diff
@@ -77,6 +85,11 @@ export const approveForParent = async ({
   const protocolFee = await zns.curvePricer.getFeeForPrice(ethers.ZeroHash, price + parentFee);
   const toApprove = price + parentFee + protocolFee;
 
+  const userBalance = await tokenContract.balanceOf(user.address);
```
JamesEarle, line 86: Considering this is solely for testing, would it be better to avoid the extra await and just give the user funds regardless?

test/ZNSRootRegistrar.test.ts
```diff
@@ -443,6 +327,56 @@ describe("ZNSRootRegistrar", () => {
         ZERO_ADDRESS_ERR
       );
     });
+
+    it("#pauseRegistration() should revert if called by address without ADMIN_ROLE", async () => {
```
JamesEarle, line 352: its unnecessarily duplicative to put `#[functionName]` at the beginning of every `it` call. For some of the test changes we've been discussing already it would be nicer to have a describe suite of unit tests per function

```typescript
describe("#pauseRegistration", () => {
  it("should revert if called by address without ADMIN_ROLE", async () => {
    ...
  });
});
```

test/ZNSRootRegistrar.test.ts
```diff
@@ -443,6 +327,56 @@ describe("ZNSRootRegistrar", () => {
...
+    it("#unpauseRegistration() should revert if called by address without ADMIN_ROLE", async () => {
+      const isAdmin = await zns.accessController.hasRole(ADMIN_ROLE, randomUser.address);
+      expect(isAdmin).to.be.false;
+
+      await expect(zns.rootRegistrar.connect(randomUser).unpauseRegistration())
+        .to.be.revertedWithCustomError(zns.accessController, AC_UNAUTHORIZED_ERR)
+        .withArgs(randomUser.address, ADMIN_ROLE);
+    });
+
+    it("#unpauseRegistration() should unpause the registration and emit #RegistrationPauseSet", async () => {
```
JamesEarle, line 386: a small preference, but I try to have success path tests all before failures, nitpick though

test/ZNSRootRegistrar.test.ts
```diff
@@ -873,6 +844,55 @@ describe("ZNSRootRegistrar", () => {
...
+        paymentConfig: {
+          token: await zns.meowToken.getAddress(),
+          beneficiary: user.address,
+        },
+      });
+
+      const domainHash = hashDomainLabel(defaultDomain);
+      const config = await zns.treasury.paymentConfigs(domainHash);
+      expect(config.token).to.eq(await zns.meowToken.getAddress());
+      expect(config.beneficiary).to.eq(user.address);
```
JamesEarle, line 907: Not sure where I would group this type of test when we rafactor, because we're confirming behaviour on contract that isn't this test file's contract, but is related to it. Probably flow, maybe integration?

test/ZNSRootRegistrar.test.ts
```diff
@@ -443,6 +327,56 @@ describe("ZNSRootRegistrar", () => {
...
+    it("#pauseRegistration() should revert if called by address without ADMIN_ROLE", async () => {
+      const isAdmin = await zns.accessController.hasRole(ADMIN_ROLE, randomUser.address);
+      expect(isAdmin).to.be.false;
+
+      await expect(zns.rootRegistrar.connect(randomUser).pauseRegistration())
+        .to.be.revertedWithCustomError(zns.accessController, AC_UNAUTHORIZED_ERR)
+        .withArgs(randomUser.address, ADMIN_ROLE);
+    });
+
+    it("#pauseRegistration() should pause the registration and emit #RegistrationPauseSet event", async () => {
```
JamesEarle, line 361: Consider also adding a few tests that show non-admin users can still use other functionality when paused

test/ZNSSubRegistrar.test.ts
```diff
@@ -141,214 +143,357 @@ describe("ZNSSubRegistrar", () => {
...
         },
-      };
+      });
+
+      const tokenId = BigInt(subHash).toString();
+      const tokenURI = await zns.domainToken.tokenURI(tokenId);
+      expect(tokenURI).to.eq(subTokenURI);
+    });
+
+    it("Can register a subdomain with characters [a-z0-9]", async () => {
```
JamesEarle: We also allow hyphens, should be `[a-z0-9-]`

contracts/registrar/ARegistrationPause.sol
Whytecrowe: in my head it is different from the regular pause we have on V1. because that one pauses all contract setters as well. this one ONLY pauses registration

test/helpers/register-setup.ts
Whytecrowe: I don't really think it slows it down that much. does it?

test/ZNSRootRegistrar.test.ts
Whytecrowe: yes, I agree. there the contexting was already present, so I decided to not do that.
However, when we refactor, I do want to do this. will see if I can fix this here as well

test/ZNSRootRegistrar.test.ts
Whytecrowe: don't really think it's necessary, tbh. since those functions don't have the modifier. it's apparent they are available

test/ZNSRootRegistrar.test.ts
Whytecrowe: I think this should be here, since it's realted to Registrar doing a proper call to set values. direct changes to payment config only are in the Treasury test. this is just registration with more params test

PR #139, opened 2025-06-04
fix: Domain Test Helper Class

Whytecrowe:


src/deploy/missions/types.ts
```diff
@@ -7,3 +11,12 @@ export interface ICurvePriceConfig {
...
 }
+
+export interface IFixedPriceConfig {
+  price : bigint;
+  feePercentage : bigint;
+}
+
+// TODO dom: refactor the types above to use these types below instead in all of the code !
+export type CurvePriceConfig = Partial<Pick<ICurvePriceConfig.CurvePriceConfigStruct, "isSet">>;
+export type FixedPriceConfig = Partial<Pick<IZNSFixedPricer.PriceConfigStruct, "isSet">>;
```
Whytecrowe: need to figure this out

test/ZNSAccessController.test.ts
```diff
@@ -6,20 +6,20 @@ import {
...
   DeployZNSParams,
-  IZNSContractsLocal,
   ZERO_ADDRESS_ERR,
 } from "./helpers";
 import { expect } from "chai";
 import { ADMIN_ROLE, DOMAIN_TOKEN_ROLE, EXECUTOR_ROLE, GOVERNOR_ROLE, REGISTRAR_ROLE } from "../src/deploy/constants";
 import { ethers } from "hardhat";
+import { IZNSContracts } from "../src/deploy/campaign/types";
 
 describe("ZNSAccessController", () => {
```
Whytecrowe, line 17: two line space always

test/ZNSAddressResolver.test.ts
```diff
@@ -53,19 +56,23 @@ describe("ZNSAddressResolver", () => {
...
-    await zns.registry.connect(deployer).addResolverType(DEFAULT_RESOLVER_TYPE, await zns.addressResolver.getAddress());
+    domain = new Domain({
+      zns,
+      domainConfig: {
+        label: "wilder",
+        owner: deployer,
+        parentHash: ethers.ZeroHash,
+        tokenOwner: deployer.address,
+        distrConfig: distrConfigEmpty,
+        priceConfig: {},
```
Whytecrowe: why is this passed as an empty object?

test/ZNSAddressResolver.test.ts
```diff
@@ -53,19 +56,23 @@ describe("ZNSAddressResolver", () => {
...
 
-    await zns.registry.connect(deployer).addResolverType(DEFAULT_RESOLVER_TYPE, await zns.addressResolver.getAddress());
+    domain = new Domain({
+      zns,
+      domainConfig: {
+        label: "wilder",
+        owner: deployer,
+        parentHash: ethers.ZeroHash,
+        tokenOwner: deployer.address,
+        distrConfig: distrConfigEmpty,
```
Whytecrowe: this should be assigned automatically in the class, so here we don't need to pass it.

test/ZNSAddressResolver.test.ts
```diff
@@ -53,19 +56,23 @@ describe("ZNSAddressResolver", () => {
...
+        priceConfig: {},
+        paymentConfig: paymentConfigEmpty,
```
Whytecrowe: same here. empty configs should be assigned as defaults by the class when not passed.

test/ZNSAddressResolver.test.ts
```diff
@@ -83,14 +90,13 @@ describe("ZNSAddressResolver", () => {
 
   it("Should get the AddressResolver", async () => { // Copy of registry tests
     // The domain exists
-    const existResolver = await zns.registry.getDomainResolver(wilderDomainHash);
+    const existResolver = await domain.getResolverAddressByLabel(domain.label);
```
Whytecrowe: why by label?

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+  get tokenId () : bigint {
+    return BigInt(this.hash);
+  }
+
+  async ownerOfHash () : Promise<string> {
+    return this.zns.registry.getDomainOwner(this.hash);
+  }
+
+  async ownerOfToken () : Promise<string> {
+    return this.zns.domainToken.ownerOf(BigInt(this.tokenId));
```
Whytecrowe: this.tokenId is already a BigInt, no need to wrap it again

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+      undefined,
+    );
+
+    const events = await this.zns.rootRegistrar.queryFilter(filter, latestBlock - 2, latestBlock);
+    const { args: { domainHash } } = events[events.length - 1];
+
+    return domainHash;
+  }
+
+  async mintAndApproveForDomain (user ?: SignerWithAddress) : Promise<ContractTransactionResponse> {
```
Whytecrowe: this should check if there's enough approval there already, and only approve if there's not enough

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+      });
+    }
+
+    this.hash = await this.getDomainHashFromEvent(executor);
+
+    return txPromise;
+  }
+
+  async revoke (executor ?: SignerWithAddress) : Promise<void> {
+    await this.zns.rootRegistrar.connect(executor ? executor : this.owner).revokeDomain(this.hash);
```
Whytecrowe: do a `return` here instead of `await`

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  async assignDomainToken (
+    to : string,
+    executor ?: SignerWithAddress
+  ) : Promise<ContractTransactionReceipt | null> {
+    const tx = await this.zns.rootRegistrar.connect(executor ? executor : this.owner).assignDomainToken(
+      this.hash,
+      to
+    );
+    return tx.wait();
```
Whytecrowe: why this has `tx.wait()` but no other methods do. either add everywhere with a configurable confirmation number or remove everywhere

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  async updateDomainRecord (
+    resolverType : string,
+    executor ?: SignerWithAddress
+  ) : Promise<void> {
+    await this.zns.registry.connect(executor ? executor : this.owner).updateDomainRecord(
```
Whytecrowe: return

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+    allowed : Array<boolean>,
+    executor ?: SignerWithAddress
+  ) : Promise<void> {
+    if (candidates.length !== allowed.length)
+      throw new Error("Domain Helper: Candidates and allowed arrays must have the same length");
+
+    await this.zns.subRegistrar.connect(executor ? executor : this.owner).updateMintlistForDomain(
+      this.hash,
+      candidates ? candidates : [this.owner.address],
+      allowed ? allowed : [true],
```
Whytecrowe: there shouldn't be default parameters here. if incorrect ones or none are passed it should just fail

Whytecrowe: also do `return` instead of await for all methods that end in a call

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+  ) : Promise<void> {
+    await this.zns.subRegistrar.connect(executor ? executor : this.owner).setDistributionConfigForDomain(
+      this.hash,
+      this.distrConfig,
+    );
+  }
+
+  async setPricerDataForDomain (
+    priceConfig : CurvePriceConfig | FixedPriceConfig,
+    pricerContract : string,
```
Whytecrowe: optional parameter.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+    executor ?: SignerWithAddress
+  ) : Promise<void> {
+    await this.zns.subRegistrar.connect(executor ? executor : this.owner).setPricerDataForDomain(
```
Whytecrowe: all methods in this class should `return`

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+  async setOwnersOperator (
+    operator : string,
+    allowed : boolean,
+    executor ?: SignerWithAddress
+  ) : Promise<void> {
+    await this.zns.registry.connect(executor ? executor : this.owner).setOwnersOperator(
+      operator,
+      allowed
+    );
+  }
```
Whytecrowe: this method is added twice. there's already one above that does the same thing

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  // TODO: delete this
+  async getResolverAddressByLabel (label : string) : Promise<string> {
+    const hash = hashDomainLabel(label);
+    return this.zns.registry.getDomainResolver(hash);
+  }
```
Whytecrowe, line 298: ?

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  // ------------------------------------------------------
+  // VALIDATION
+  // ------------------------------------------------------
+  async validateDomainHash (expectedHash : string) : Promise<void> {
+
+  }
```
Whytecrowe: ?

test/ZNSRootRegistrar.test.ts
```diff
@@ -57,26 +55,30 @@ import { runZnsCampaign } from "../src/deploy/zns-campaign";
...
-import { ICurvePriceConfig  } from "../src/deploy/missions/types";
+import { ICurvePriceConfig, IFixedPriceConfig  } from "../src/deploy/missions/types";
 import { IZNSContracts } from "../src/deploy/campaign/types";
+import Domain from "./helpers/domain/domain";
 
 require("@nomicfoundation/hardhat-chai-matchers");
 
 
-describe("ZNSRootRegistrar", () => {
+describe.only("ZNSRootRegistrar", () => {
```
Whytecrowe: remove .only

test/ZNSRootRegistrar.test.ts
```diff
@@ -116,41 +118,11 @@ describe("ZNSRootRegistrar", () => {
 
   afterEach(async () => {
     await mongoAdapter.dropDB();
+
+    domain = undefined as unknown as Domain;
```
Whytecrowe: why is this here and what is it?

test/ZNSRootRegistrar.test.ts
```diff
@@ -613,24 +507,18 @@ describe("ZNSRootRegistrar", () => {
...
+          distrConfig,
         },
       });
 
-      const domainHash = await getDomainHashFromEvent({
-        zns,
-        user,
-      });
+      await domain.registerAndValidateDomain();
+      const domainHash = await domain.getDomainHashFromEvent(user);
```
Whytecrowe: `domain.hash` should be here instead.

test/ZNSRootRegistrar.test.ts
```diff
@@ -720,16 +599,17 @@ describe("ZNSRootRegistrar", () => {
...
+          tokenURI,
+        },
       });
+      await domain.register();
 
-      directPaymentDomainHash = await getDomainHashFromEvent({
-        zns,
-        user,
-      });
+      directPaymentDomainHash = await domain.getDomainHashFromEvent(user);
```
Whytecrowe: domain.hash should be enough

Whytecrowe (commented):
comments

test/helpers/domain/domain.ts
MichaelKorchagin: It does, but it mints and approves the full ```totalPrice``` when ```userBalance < totalPrice```. Improved it to ```mint(totalPrice - userBalance)``` and ```approve(totalPrice)```

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+    this.zns = zns;
+
+    this.owner = domainConfig.owner;
+    this.parentHash = domainConfig.parentHash || hre.ethers.ZeroHash;
+    this.isRoot = this.parentHash === hre.ethers.ZeroHash;
+    this.label = domainConfig.label;
+    this.tokenOwner = domainConfig.tokenOwner || hre.ethers.ZeroAddress;
+    this.distrConfig = domainConfig.distrConfig || distrConfigEmpty;
+
+    switch (this.distrConfig.pricerContract) {
```
Whytecrowe: this is a limited approach. what if a specific priceConfig has already been passed inside `domainConfig` argument?
How did you handle cases of specific configs passed to this class?

This flow here should only be launche IF `domainConfig.priceConfig` is non-existent, otherwise it should just use that one. This would allow this to be directly and easily configurable at construction time.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+      label,
+      parentHash,
+      distrConfig,
+      paymentConfig,
+      tokenURI,
+      tokenOwner,
+      domainAddress,
+    } = this;
+
+    if (!distrConfig.priceConfig) {
```
Whytecrowe: is this even possible if you already did the whole flow to set it in the constructor? this case should not exist at this point. priceConfig should already be correctly assigned in the constructor. if so, no IF block is needed here

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+      distrConfig.priceConfig = encodePriceConfig(this.priceConfig);
+    }
+
+    let txPromise : ContractTransactionResponse;
+
+    // mint and approve strict amount of tokens for domain registration
+    await this.mintAndApproveForDomain(executor);
+
+    if (this.isRoot) {
+      txPromise = await zns.rootRegistrar.connect(executor ? executor : owner).registerRootDomain({
```
Whytecrowe: No reason to determine the caller here in every call.

Just add a line above everything like:

const caller = executor ? executor : owner.

And then use this `caller` var everywhere below.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+        label,
+        domainAddress: hre.ethers.isAddress(domainAddress) ? domainAddress : owner.address,
+        tokenOwner,
+        tokenURI,
+        distrConfig,
+        paymentConfig,
+      });
+    }
+
+    this.hash = await this.getDomainHashFromEvent(executor);
```
Whytecrowe: this same `caller` var should be used here as well. this is currently incorrect and will not work if you don't pass the `executor` parameter, which is not needed in 99% of the cases.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+    executor ?: SignerWithAddress
+  ) : Promise<ContractTransactionResponse> {
+    return this.zns.rootRegistrar.connect(executor ? executor : this.owner).assignDomainToken(
+      this.hash,
+      to
+    );
+  }
+
+  async updateDomainRecord (
+    resolverType : string,
```
Whytecrowe: why is this only accepts `resolverType` and not `owner`? this method on the contract changes both. `owner` should be a parameter as well and both `owner` and `resolverType`params should be optional, and then use the proper default value if one is not passed. that's how the contract works.

Do it the same way you did `setPricerDataForDomain` method.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+    executor ?: SignerWithAddress
+  ) : Promise<ContractTransactionResponse> {
+    return this.zns.registry.connect(executor ? executor : this.owner).updateDomainRecord(
+      this.hash,
+      this.owner,
```
Whytecrowe: needs to be a parameter and only passed as `this.owner` here if the argument was not provided

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+    return this.zns.subRegistrar.connect(executor ? executor : this.owner).updateMintlistForDomain(
+      this.hash,
+      candidates,
+      allowed,
+    );
+  }
+
+  // ------------------------------------------------------
+  // SETTERS
+  // ------------------------------------------------------
```
Whytecrowe: incorrect comment, since methods above are setters as well :)
Can call this something more appropriate or just remove the comment altogether

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+        BigInt(this.hash),
+        this.tokenURI,
+        executor ? executor.address : this.owner.address,
+        executor ? executor.address : this.owner.address,
+        this.domainAddress
+      );
+
+    // check domain existence with registry
+    const record = await this.zns.registry.getDomainRecord(this.hash);
+    const resolverAddress = await this.zns.registry.getDomainResolver(this.hash);
```
Whytecrowe: there is no need to call this as well, since it checks the same exact mapping `getDomainRecord()` returns. Comparing these 2 values checks the same variable against itself.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    expect(
+      await this.zns.registry.getDomainOwner(this.hash)
+    ).to.equal(executor ? executor.address : this.owner.address);
+    expect(record.owner).to.equal(executor ? executor.address : this.owner.address);
+    expect(record.resolver).to.equal(resolverAddress);
```
Whytecrowe: this check is redundant and compares the same thing to itself. instead we can compare `record.resolver` to one of the only 2 resolvers we have. it should match one or the other or zero address if hasn't been set during registration

test/helpers/index.ts
```diff
@@ -13,3 +13,4 @@ export { EXECUTOR_ROLE } from "../../src/deploy/constants";
 export { REGISTRAR_ROLE } from "../../src/deploy/constants";
 export { ADMIN_ROLE } from "../../src/deploy/constants";
 export { GOVERNOR_ROLE } from "../../src/deploy/constants";
+export { IFixedPriceConfig } from "../../src/deploy/missions/types";
```
Whytecrowe: this seems incorrect to me and should not be here.

test/ZNSRootRegistrar.test.ts
```diff
@@ -885,89 +756,67 @@ describe("ZNSRootRegistrar", () => {
...
+      await domain.register();
 
-      const tokenId = BigInt(
-        await getDomainHashFromEvent({
-          zns,
-          user,
-        })
-      );
-      const owner = await zns.domainToken.ownerOf(tokenId);
+      const owner = await zns.domainToken.ownerOf(await domain.tokenId);
```
Whytecrowe: await is redundant here for `domain.tokenId`

test/ZNSRootRegistrar.test.ts
```diff
@@ -885,89 +756,67 @@ describe("ZNSRootRegistrar", () => {
...
 
     it("Resolves the correct address from the domain", async () => {
-      await defaultRootRegistration({
-        user,
-        zns,
-        domainName: defaultDomain,
-        domainContent: await zns.rootRegistrar.getAddress(),
-      });
-      const domainHash = await getDomainHashFromEvent({
+      domain = new Domain({
```
Whytecrowe: there's no need to keep reusing the same variable in all tests.
If you create it with `const` for every test, that should actually use less memory, since you don't need to propagate redundant data between individual closures of each test, so it should be wiped between `it()` runs by the garbage collector. but this way it just holds data there across the whole test file, and we actually do not need it after the test has finished.

Only use large scope variable if you actually need its data across multiple closures, otherwise, creating a new var for each test is better.

Apart from that, with declarations and usage like this, there's a possibility that state from the previous tests will bleed into other tests, and can create problems if we decide to run tests in parallel. So this is bad practice.

test/ZNSRootRegistrar.test.ts
```diff
@@ -1012,124 +861,140 @@ describe("ZNSRootRegistrar", () => {
...
         zns,
-        domainName: defaultDomain,
-        tokenOwner: ethers.ZeroAddress,
-        tokenURI,
-        distrConfig,
+        domainConfig: {
+          owner: user,
+          label: defaultDomain,
+          distrConfig,
+          paymentConfig: {
```
Whytecrowe, line 897: don't need to pass this, since it will assign empty config in `Domain` constructor.

test/ZNSRootRegistrar.test.ts
```diff
@@ -1371,27 +1239,20 @@ describe("ZNSRootRegistrar", () => {
 
       const asBytes = encodePriceConfig(newConfig);
 
-      await zns.subRegistrar.connect(user).setPricerDataForDomain(
-        domainHash,
-        asBytes,
-        zns.fixedPricer.target,
+      await domain.setPricerDataForDomain(
+        newConfig,
+        zns.fixedPricer.target as string,
```
Whytecrowe: this type casting should not be needed. just make the type for this parameter in the method wider so it supports both, the `target` prop type and regular strings as well

test/ZNSRootRegistrar.test.ts
```diff
@@ -1429,11 +1290,16 @@ describe("ZNSRootRegistrar", () => {
...
         zns,
-        user,
+        domainConfig: {
+          owner: user,
+          label: defaultDomain,
+        },
       });
+      await domain.register();
+
+      const domainHash = await domain.getDomainHashFromEvent(user);
```
Whytecrowe: why this call? shouldn't it already be inside `domain.hash`?

test/ZNSRootRegistrar.test.ts
```diff
@@ -1468,64 +1333,78 @@ describe("ZNSRootRegistrar", () => {
...
         zns,
-        user: deployer,
+        domainConfig: {
+          owner: deployer,
+          label: defaultDomain,
+        },
       });
-      const owner = await zns.registry.connect(user).getDomainOwner(parentDomainHash);
+      await domain.register();
+      const parentDomainHash = await domain.getDomainHashFromEvent(deployer);
```
Whytecrowe: same here. no need for this call. just pass `domain.hash` where it's needed

test/ZNSStringResolver.test.ts
```diff
@@ -23,10 +23,12 @@ import {
 import { DeployCampaign, MongoDBAdapter } from "@zero-tech/zdc";
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { getConfig } from "../src/deploy/campaign/get-config";
-import { IZNSContractsLocal } from "./helpers/types";
+import Domain from "./helpers/domain/domain";
 
 
 describe("ZNSStringResolver", () => {
+  let domain : Domain;
```
Whytecrowe, line 30: same here. no need for this large scope var in any test. just declare them inside each it()

Whytecrowe:
Please, address comments

src/deploy/missions/types.ts
```diff
@@ -1,4 +1,4 @@
-
+// eslint-disable-next-line no-redeclare
```
JamesEarle: Why is this here? this interface isn't redeclared anywhere in this file? Am I missing something?

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+
+  async ownerOfHash () : Promise<string> {
+    return this.zns.registry.getDomainOwner(this.hash);
+  }
+
+  async ownerOfToken () : Promise<string> {
+    return this.zns.domainToken.ownerOf(this.tokenId);
+  }
+
+  async getDomainHashFromEvent (domainOwner ?: SignerWithAddress) : Promise<string> {
```
JamesEarle: This function should be declared `private`, the class should only expose the domain hash through the `hash` property

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+    if (!distrConfig.priceConfig) {
+      distrConfig.priceConfig = encodePriceConfig(this.priceConfig);
+    }
+
+    let txPromise : ContractTransactionResponse;
+
+    // mint and approve strict amount of tokens for domain registration
+    await this.mintAndApproveForDomain(executor);
+
+    if (this.isRoot) {
```
JamesEarle: Why bother doing all the deconstruction of `this` above on L110-L120 when you just access using `this.[prop_name]` anyways? Pick one and be consistent

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+  async mintAndApproveForDomain (user ?: SignerWithAddress) : Promise<ContractTransactionResponse> {
+    return (fundApprove({
+      zns: this.zns,
+      parentHash: this.parentHash,
+      user: user ? user : this.owner,
+      domainLabel: this.label,
+    })) as Promise<ContractTransactionResponse>;
+  }
+
+  async register (executor ?: SignerWithAddress) : Promise<ContractTransactionResponse> {
```
JamesEarle, line 136: The presence of this function feels weird. So the expected flow is to
 - Create a new `Domain` class instance before the domain actually exists with `const myDomain = new Domain(zns, domainConfig)`
 - then call `domain.register` through this function?

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  async setPricerDataForDomain (
+    priceConfig ?: ICurvePriceConfig | IFixedPriceConfig,
+    pricerContract ?: string,
+    executor ?: SignerWithAddress
+  ) : Promise<ContractTransactionResponse> {
+    return this.zns.subRegistrar.connect(executor ? executor : this.owner).setPricerDataForDomain(
+      this.hash,
+      priceConfig ? encodePriceConfig(priceConfig) : encodePriceConfig(this.priceConfig),
```
JamesEarle: This would throw an error if the incoming `priceConfig` param is `undefined` and the default case above is hit in your switch statement where `priceConfig === {}`

Resolve the `priceConfig` to use before this return statement, and if the incoming param is `undefined` and `this.priceConfig` is `{}`

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+      this.hash,
+      candidates,
+      allowed,
+    );
+  }
+
+  // ------------------------------------------------------
+  // SETTERS
+  // ------------------------------------------------------
+  async setOwnersOperator (
```
JamesEarle, line 223: There should be additional functionality on this class that enables users to also check if an address is an operator for the domain, view the operators of a domain, etc.

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+    operator : string,
+    allowed : boolean,
+    executor ?: SignerWithAddress
+  ) : Promise<ContractTransactionResponse> {
+    return this.zns.registry.connect(executor ? executor : this.owner).setOwnersOperator(operator, allowed);
+  }
+
+  async setDistributionConfigForDomain (
```
JamesEarle, line 231: This should accept a `distrConfig` parameter that you want to set, not just use the one that was set by the constructor because that one can possibly be empty

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,287 @@
...
+  ) : Promise<ContractTransactionResponse> {
+    return this.zns.subRegistrar.connect(executor ? executor : this.owner).setPricerDataForDomain(
+      this.hash,
+      priceConfig ? encodePriceConfig(priceConfig) : encodePriceConfig(this.priceConfig),
+      pricerContract ? pricerContract : this.distrConfig.pricerContract
+    );
+  }
+
+  async setPaymentTypeForDomain (
+    paymentType : bigint,
```
JamesEarle, line 271: It's technically possible to have this be any number which would invalidate the config, and because we don't do on chain validation of the value itself it would allow setting it. 

Add validation here to confirm the number is within the range of expected numbers. Same for `setAccessTypeForDomain` below

test/helpers/domain/domain.ts
Whytecrowe: correct. why weird? we create instance of the domain with all the initial data, then we need an async call to register it.

test/helpers/domain/domain.ts
Whytecrowe: good catch. otherwise this function is kind of useless.
should we also set that argument as this.distrConfig after this method is called you think?

test/helpers/domain/domain.ts
MichaelKorchagin: I was the same point :))
I forgot to run ```domain.register()`` several times, and each time I was surprised why a simple process didn't work.
I still don't see any cases where we need a Domain-class instance without having an existing domain. 
The only good point of this is that it is more convenient to work with. So we don't need async contructors to run registration in it and so on.

test/helpers/domain/domain.ts
MichaelKorchagin: I install it for the whole class after. Thank you!

test/helpers/domain/domain.ts
```diff
@@ -0,0 +1,349 @@
...
+    this.owner = domainConfig.owner;
+    this.label = domainConfig.label;
+
+    // setting up all optional params
+    this.parentHash = domainConfig.parentHash || hre.ethers.ZeroHash;
+    this.isRoot = this.parentHash === hre.ethers.ZeroHash;
+    this.tokenOwner = domainConfig.tokenOwner || this.owner.address;
+    this.distrConfig = domainConfig.distrConfig || distrConfigEmpty;
+
+    if (!domainConfig.priceConfig) {
```
Whytecrowe, line 66: call this prop `priceConfigDecoded` or `priceConfigDec`

PR #149, opened 2025-07-22
Audit fix / "Admin can replace the Governor role"

MichaelKorchagin:


src/utils/migration/helpers.ts
```diff
@@ -107,7 +107,7 @@ const createBatchesEOA = (
     // "label" is parameter name for subdomain registration, "name" for root domains
     // but "label" is used for both in the subgraph
     if (domain.parentHash !== ZeroHash) {
-      const { name, ...rest } = domainArg;
+      const { ...rest } = domainArg;
```
Whytecrowe: wrong change! here in this code we specifically desctusture `name` to remove it from the object data, now that it's removed, `...rest` will include the `name` prop, which is incorrect.
This should've just been comment to silence the linter, which has already been applied in another PR.

src/utils/migration/helpers.ts
```diff
@@ -177,7 +177,7 @@ const createBatchesSafe = (
       }
 
       // remove `name` from args, leave rest of args in `rest`
-      const { name, ...rest } = args;
+      const { ...rest } = args;
```
Whytecrowe: same thing here. `name` should NOT be removed from destructuring

PR #150, opened 2025-07-23
Fix - A Single token owner can block all transfers in batch

JamesEarle:
Address the audit issue created by  the possibility of `transfers` beings sent to contracts that may not implement `onERC721Received`, which would cause a failing tx in the batch. As batches are executed by the Safe in a single tx, this would fail the entire batch not just this transactions.

To address this, we first get the bytecode for every potential transfer and then check the length. EOAs will have length of 2, just of `0x` but contracts will have more. We could additionally try to find out of the contract supports `onERC721Received` with something like `supportsInterface` however this isn't reliable and we've agreed it's enough to solve the problem by flagging any non-EOA transfer recipient instead.

 - [x] A Single token owner can block all transfers in batch

### DIFF
--- a/src/utils/migration/helpers.ts
+++ b/src/utils/migration/helpers.ts
@@ -315,21 +315,17 @@ export const createTransfers = async (
       operation: OperationType.Call,
     };
 
-    const bytecode = await hre.ethers.provider.getCode(domain.domainToken.owner.id);
-
-    if (bytecode.length > 2) {
-      try {
-        // If destination address is EOA or contract that implements `onERC721Received`
-        // this should pass. Otherwise, we mark the transfer as a failure to avoid failing
-        // the entire batch later
-        await domainToken["safeTransferFrom(address,address,uint256)"].estimateGas(
-          safeAddress,
-          domain.domainToken.owner.id,
-          domain.tokenId
-        );
-      } catch (e) {
-        failedTransfers.push(domain);
-      }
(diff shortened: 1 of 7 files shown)


src/utils/migration/helpers.ts
```diff
@@ -298,18 +307,38 @@ export const createTransfers = (
...
+            domain.domainToken.owner.id,
+            domain.tokenId
+          );
+        } catch (e) {
+          failedTransfers.push(domain);
+        }
+      }
+
+      // The `to` address must be the contract the multisig will call,
+      // not the multisig itself
```
Whytecrowe: this comment needs moving above

src/utils/migration/helpers.ts
```diff
@@ -298,18 +307,38 @@ export const createTransfers = (
...
+        to: await domainToken.getAddress(),
         value: "0",
         data: transferEncoding,
         operation: OperationType.Call,
-      });
+      };
+
+      const bytecode = await hre.ethers.provider.getCode(domain.domainToken.owner.id);
+
+      if (bytecode.length > 2) {
```
Whytecrowe: we should estimate gas for every call, not only when receiver is a contract. this would let us cover more potential error and not just safeTransfer specific ones

Whytecrowe (changes requested):
please address comments

PR #152, opened 2025-07-28
[AUDIT] Revoked domains remain locked under Safe ownership post migration

MichaelKorchagin:


src/utils/migration/02_registration.ts
```diff
@@ -118,6 +118,22 @@ const main = async () => {
     SUB_COLL_NAME
   ).find().sort({ depth: 1, _id: 1 }).toArray() as unknown as Array<Domain>;
 
+  const rootsToRevoke = await client.collection(ROOT_COLL_NAME).find(
```
JamesEarle: You don't need to call to the db again to get this data. We just called on the lines above this, so all you would need to do is filter the data you've already gotten instead of making new queries.

src/utils/migration/02_registration.ts
```diff
@@ -212,15 +228,41 @@ const main = async () => {
     // Create and propose the batch transactions
     await safeKit.createProposeBatches(transferTxs, 100);
     break;
+  case "revoke":
+    // Revoke domains that are owned by the Safe
+    const {
+      revokeBatches,
+      failedRevokes,
+    } = await createRevokeBatches(
+      domainsToRevoke,
```
JamesEarle, line 250: Like the above comment, you could instead just filter on the `rootDomains` and `subdomains` variables you already have, instead of calling to the DB again.

`const rootsToRevoke = rootDomains.filter((domain) => domain.isRevoked === true)` is sufficient. And because we're checking a boolean value already, we can just do `((domain) => domain.isRevoked)`

The array you pass to here would be doing the above call to `.filter` on both rootDomains and subdomains arrays

src/utils/migration/02_registration.ts
```diff
@@ -93,22 +93,22 @@ const main = async () => {
...
+  // const safeKit = await SafeKit.init(config);
 
-  // If admin given is not a Safe owner, fail early
-  if (!await safeKit.isOwner(migrationAdmin.address)) {
-    throw new Error(
-      `Migration admin ${migrationAdmin.address} is not a Safe owner. Ensure the admin address is added as a Safe owner`
-    );
-  }
+  // // If admin given is not a Safe owner, fail early
+  // if (!await safeKit.isOwner(migrationAdmin.address)) {
```
JamesEarle: Uncomment this. If you want some changes for local testing add to check the `hre.network.name === "hardhat"`

src/utils/migration/02_registration.ts
```diff
@@ -93,22 +93,22 @@ const main = async () => {
   // https://docs.safe.global/advanced/smart-account-supported-networks?service=Transaction+Service&service=Safe%7BCore%7D+SDK
 
   // Setup the SafeKit
-  const safeKit = await SafeKit.init(config);
+  // const safeKit = await SafeKit.init(config);
```
JamesEarle: Similarly uncomment this

src/utils/migration/02_registration.ts
```diff
@@ -212,15 +228,41 @@ const main = async () => {
...
 };
 
-main().catch(error => {
-  const logger = getZnsLogger();
-  logger.error("Migration script failed:", error);
-  process.exitCode = 1;
-}).finally(() => {
-  process.exit(0);
-});
+// main().catch(error => {
```
JamesEarle: I get that you commented this out to import it as a module in your tests, but this is not an executable script if this remains commented. Please uncomment this

src/utils/migration/02_registration.ts
```diff
@@ -57,7 +57,7 @@ import { getZnsLogger } from "../../deploy/get-logger";
  * - Executing more than ~20 transactions sequentially is not recommended
  * - Manual gas estimation is used for complex transaction accuracy
  */
-const main = async () => {
+export const main = async () => {
```
Whytecrowe: this export should not be added.
NodeJS works un such way that if you have a root level call in the file and then you import something from this file, this script will run every time you run anything.
If you need a function to be exportable, it should be in a different file than a root level script, otherwise it could lead to all sorts of issues.

src/utils/migration/02_registration.ts
```diff
@@ -118,6 +118,22 @@ const main = async () => {
     SUB_COLL_NAME
   ).find().sort({ depth: 1, _id: 1 }).toArray() as unknown as Array<Domain>;
 
+  const rootsToRevoke = await client.collection(ROOT_COLL_NAME).find(
+    {
+      isRevoked: true,
+      // "owner.id": safeAddress,
```
Whytecrowe: this should probably be removed

src/utils/migration/02_registration.ts
```diff
@@ -118,6 +118,22 @@ const main = async () => {
...
+    }
+  ).toArray();
+
+  const subsToRevoke = await client.collection(SUB_COLL_NAME).find(
+    {
+      isRevoked: true,
+      // "owner.id": safeAddress,
```
Whytecrowe: same here

src/utils/migration/02_registration.ts
Whytecrowe: yes. use the above arrays that have already been pulled from the DB

src/utils/migration/02_registration.ts
```diff
@@ -212,15 +228,41 @@ const main = async () => {
...
+    } = await createRevokeBatches(
+      domainsToRevoke,
+      zns.rootRegistrar,
+    );
+
+    // if (failedRevokes.length > 0) {
+    //   const result = await client.collection(INVALID_REVOKES_COLL_NAME).insertMany(failedRevokes);
+    //   const diff = failedRevokes.length - result.insertedCount;
+    //   if (diff > 0) {
+    //     throw new Error(`Failed to insert ${diff} failed domain transfers`);
```
JamesEarle: `...failed domain transfers);` update the error message

src/utils/migration/02_registration.ts
```diff
@@ -212,15 +228,41 @@ const main = async () => {
...
+    // Revoke domains that are owned by the Safe
+    const {
+      revokeBatches,
+      failedRevokes,
+    } = await createRevokeBatches(
+      domainsToRevoke,
+      zns.rootRegistrar,
+    );
+
+    // if (failedRevokes.length > 0) {
```
JamesEarle: uncomment this safety check

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      admins: [deployer.address, admin.address],
+    });
+
+    const campaign = await runZnsCampaign({
+      config,
+    });
+
+    zns = campaign.state.contracts;
+
+    await zns.accessController.connect(deployer).grantRole(DOMAIN_TOKEN_ROLE, await zns.domainToken.getAddress());
```
Whytecrowe: this should already be done during the campaign in `DomainTokenDM.postDeploy()`. This call is doing the same thing again.

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    // Give funds to user
+    await zns.meowToken.connect(deployer).approve(await zns.treasury.getAddress(), hre.ethers.MaxUint256);
+    await zns.meowToken.mint(deployer.address, userBalanceInitial);
+
+    const domainsDbClient = await connectToDb();
+
+    rootsToRevoke = await domainsDbClient.collection(ROOT_COLL_NAME).find(
+      {
+        isRevoked: true,
+        // "owner.id": safeAddress,
```
Whytecrowe: remove

src/utils/migration/constants.ts
```diff
@@ -4,6 +4,7 @@ export const ROOT_COLL_NAME = process.env.MONGO_DB_ROOT_COLL_NAME || "root-domai
 export const SUB_COLL_NAME = process.env.MONGO_DB_SUB_COLL_NAME || "subdomains";
 export const INVALID_COLL_NAME = process.env.MONGO_DB_INVALID_COLL_NAME || "invalid-domains";
 export const INVALID_TX_COLL_NAME = process.env.MONGO_DB_INVALID_COLL_NAME || "invalid-transactions";
+export const INVALID_REVOKES_COLL_NAME = process.env.MONGO_DB_INVALID_COLL_NAME || "invalid-revokes";
```
JamesEarle: make this check a different `process.env` var first, it shouldnt be looking for the same key as the above line

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    //   await zns.subRegistrar.connect(deployer).registerSubdomain({
+    //     parentHash: domain.parentHash,
+    //     label: domain.label,
+    //     domainAddress: domain.address,
+    //     tokenOwner: domain.owner.id,
+    //     tokenURI: domain.tokenURI,
+    //     distrConfig: distrConfigEmpty,
+    //     paymentConfig: paymentConfigEmpty,
+    //   });
+    // }
```
Whytecrowe: ???

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  it("run", async () => {
+    const {
+      revokeBatches,
+      failedRevokes,
+    } = await main();
```
Whytecrowe: if you use functions outside of the script by importing, change their name from `main` to something that actually tells us what this function is.
And, as I said earlier in another comment - if a function needs to be exportable, split it from the script to a separate file and rename.
`main()` functions should ONLY be inside a single script file and not used outside of that file.

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const tx = await deployer.sendTransaction({
+      to: zns.rootRegistrar.getAddress(),
+      data: hre.ethers.concat(
+        revokeBatches.map(tx => tx.data)
+      ),
+    });
```
Whytecrowe: does it actually work? I don't see any checks here that any domains have been revoked

src/utils/migration/helpers.ts
```diff
@@ -279,6 +280,49 @@ export const LogExecution = (target : any, propertyKey : string, descriptor : Pr
...
+      operation: OperationType.Call,
+    };
+
+    try {
+      await rootRegistrar.revokeDomain.estimateGas(domain.id);
+    } catch (e) {
+      failedRevokes.push(domain);
+    }
+
+    revokeBatches.push(tx);
```
Whytecrowe: this should be in the `try` block, since when you catch the error, it does NOT break the loop, so it keeps going with other code and continue through the array.

Here you find the failed tx, but since the loop is not broken you still add the invalid tx to the `revokeBatches` array, which would fail the batch.

it instead needs to only push to this array if the error did NOT happen, so moving it inside the `try` block is how we do that.

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    userBalanceInitial = hre.ethers.MaxInt256;
+
+    // Give funds to user
+    await zns.meowToken.connect(deployer).approve(await zns.treasury.getAddress(), hre.ethers.MaxUint256);
+    await zns.meowToken.mint(deployer.address, userBalanceInitial);
+
+    const domainsDbClient = await connectToDb();
+
+    rootsToRevoke = await domainsDbClient.collection(ROOT_COLL_NAME).find(
```
Whytecrowe: real environment DBs should NOT be called in a test. only localhost DBs that are filled by tests themselves.

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      hre.ethers.MaxUint256
+    );
+
+    userBalanceInitial = hre.ethers.MaxInt256;
+
+    // Give funds to user
+    await zns.meowToken.connect(deployer).approve(await zns.treasury.getAddress(), hre.ethers.MaxUint256);
+    await zns.meowToken.mint(deployer.address, userBalanceInitial);
+
+    const domainsDbClient = await connectToDb();
```
JamesEarle: Don't use a real database connection in a unit test. Always mock other services you need in a unit test like this

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        // "owner.id": safeAddress,
+      }
+    ).toArray();
+
+    subsToRevoke = await domainsDbClient.collection(SUB_COLL_NAME).find(
+      {
+        isRevoked: true,
+        // "owner.id": safeAddress,
+      }
+    ).toArray();
```
Whytecrowe: same here

test/DomainMigration.test.ts
```diff
@@ -0,0 +1,122 @@
...
+let userBalanceInitial : bigint;
+
+let zns : IZNSContractsLocal;
+let mongoAdapter : MongoDBAdapter;
+
+let domains : Array<Domain>;
+let rootsToRevoke;
+let subsToRevoke;
+
+describe.only("Domain Migration", () => {
```
JamesEarle: Remove `.only`

